PR #21, opened 2019-02-25
Feature/instance configurator

### LINKED ISSUE #7: Implement instances configuration
Done:

- [x] Add dependency to d2-ui-components
- [x] General Layout
- [x] Integrate listing table
- [x] Create new instance in global dataStore
- [x] Add Search textfield for filtering
- [x] Contextual options (Delete, Edit, Clone)
- [x] Change wizard to single page GUI
- [x] Change 'Endpoint' name to 'URL endpoint'
- [x] Implement delete
- [x] Implement tests with dev2 8083
- [x] Move description below server name.

Look and Feel:
- [x] Add spacing between Save and Cancel buttons
- [x] Save colour should be different
- [x] A bit more margin on the left side
- [x] Textfield should be full width
- [x] Fonts are slightly different than DHIS2, right? Can you have a look at it?

To-Do:
- [x] Add title and back button in all pages
- [x] Validation
   - [x] Pair URL-Username should be unique
   - [x] Check if connection is successful -> Show a different error for CORS and not found error (SEE COMMENT BELOW)
- [x] Confirm before delete with a dialog
- [x] Help button in instance-form-builder Heading, at the right of the title.
- [x] Create id by index for instances list
- [x] [Document](https://docs.google.com/document/d/1Z67aZGmoQF7tnmWSs-lpUwzglpqKQlT4wH7cwOMxx1I/edit) data store data model
- [ ] Implement [password encryption](https://www.npmjs.com/package/cryptr). Key should be read from a properties file - not hosted in github.

### LINKED ISSUE #22: Add edit/remove actions to Instance Configurator
Remove the current TODOs and implement edit/remove functionalities

SferaDev:
### :pushpin: References

* **Issue:** Closes #22

* **Issue:** Work on #7 

### :memo: Implementation

- Added TypeScript linter rules to be fully compliant

- Updated header bar to the new @dhis2/ui component

- New Instance Configurator based on d2-ui-components ObjectsTable

- New Instance Wizard based on maintenance app with built-in validation rules (isEmpty/isUrl)

- Data Store database-like implementation

- Edit/Remove existing instance from the ObjectsTable

### 👨‍💻 Work In Progress (In future Pull Requests)

- See updated card #7 

### :art: Screenshots

image

image

image

- DataStore model connector is not written in typescript, I was not sure about how to implement generic objects without breaking the no-explicit-any rule

- Pagination, next/prev doesn't work. @tokland Could you take a look please?

### :bookmark_tabs: Others

- To make selection work, I've noticed that we use "id" hardcoded on ObjectsTable, we should be able to override to a prop. Here makes sense "server" or "name".

- We have initialSorting of ObjectsTable as required, I think we could easily make it optional and in d2-ui-components assume firstColumn/asc is the default order.

- There's no way (or at least I haven't found a way) to refresh the Objects table. I have a work-around to solve that, but it should be improved in d2-ui-components.


src/models/dataStore.js
```diff
@@ -0,0 +1,69 @@
...
+        const dataStore = await d2.dataStore.create(dataStoreNamespace);
+        await dataStore.set(dataStoreKey, defaultValue);
+        return defaultValue;
+    } else {
+        const dataStore = await d2.dataStore.get(dataStoreNamespace);
+        return await dataStore.get(dataStoreKey);
+    }
+}
+
+async function saveDataStore(d2, dataStoreKey, newValue) {
```
tokland: getDataStore / saveDataStore -> same code except last line, maybe abstract common code to function.

SferaDev:
That strange behavior is because the first column is the id, and we do not have yet a validation for unique identifiers.

I've updated the card #7 with everything so we do not get lost.

@tokland Can merge the PR and I'll continue working on the issues of the card.

adrianq:
To sum up:
- All d2-ui-components related tasks that are not a high priority atm can be found here: https://github.com/EyeSeeTea/d2-ui-components/issues/8 We will use weekly dev meeting to discuss them
- @tokland it would be great if you can have a look at the pagination. This is the only real bug in d2-ui-components
- @SferaDev will create another PR with all the comments and missing features that can be found in https://github.com/EyeSeeTea/metadata-synchronization/issues/7

tokland:
I'll check the pagination. Regarding the "check button", just mention that we have `DialogHandler` in d2-ui-component which may be useful.

tokland:
@SferaDev Pagination fixed at https://github.com/EyeSeeTea/d2-ui-components/pull/11, will publish when merged.

Issue #8, opened 2019-02-12
Implement OU Sync


adrianq:
First Phase
- [x] General Layout: We will object table for now
- [x] Integrate filters: Text Search Filter
- [x] Sync Logic: 
   - Sync organisation units, organisation unit groups and organisation unit group sets and their attributes. If they don't exist we need to create it
   - Rules: If the parent OU does not exist on the slave we will not sync the OU. We will override any OU children.
   - importStrategy: CREATE_AND_UPDATE
   - mergeMode: MERGE or REPLACE
    - Exclusion rules for OU: users, userGroupAccesses, userAccesses, programs, 
dataSets, legendSets
    - Exclusion rules for OU Group: users, legendSets
    - Exclusion rules for OU Group Sets: dimensionalItemObject, users, legendSets

Second phase
- [x] Add a mechanism to come back to the home page and display the title of the page

adrianq:
Some ideas from today meeting:

**How to model exclude and include rules**
excludeRules = [legendSet, users, userGroups, organisationUnitGroups.userGroups ...]
includeRules = [ attributes, organisationUnitGroups, organisationUnitGroups.attributes, organisationUnitGroups.organisationUnitGroupSets, organisationUnitGroups.organisationUnitGroupSets.attributes]

**Data structure to be stored in the data store, sync rules, etc**
dataStructure = {
	metadataType: '',
	ids: [],
	targetInstances: [],
	originInstance: id
}

SferaDev:
**Data structure**
```js
export interface SynchronizationBuilder {
    targetInstances: string[];
    originInstance: string;
    metadata: {
        [metadataType: string]: string[];
    };
}

const builderExample: SynchronizationBuilder = {
    originInstance: "instanceId1",
    targetInstances: ["instanceId2"],
    metadata: [
        {
            metadataType1: ["myId1"],
            metadataType2: ["myId2"],
        }
    ]
};
```

**Configuration details**

importMode: COMMIT when importing, VALIDATE to check for errors
identifier: AUTO (will try for uid if not it will default to code)
importReportMode: FULL (will log debug and errors)
preheatMode: Seems to affect performance by preloading metadata before importing
importStrategy: CREATE_AND_UPDATE
atomicMode: Discuss
mergeMode: REPLACE

**Atomic Mode**

[...]

PR #25, opened 2019-03-12
feature/d2ModelFactory: Allow to reference classes dynamically

SferaDev:
### :pushpin: References

* **Issue:** Work #8 

### :memo: Implementation

- Create a factory method that dynamically references our model classes by d2 internal name (string).

- If we do not have a given model class, create a default one

### :bookmark_tabs: Others

- In the synchronization logic we use our internal model classes to extract metadata

- To abstract and translate d2 models to our own models we use this factory method

```js
const myClass = d2ModelFactory('categoryCombo');
// Try to read static class properties
console.log({metadataType: myClass.metadataType});
// Try to execute static class methods
console.log(myClass.getD2Model(d2));
```

### DIFF
--- a/src/models/d2Model.ts
+++ b/src/models/d2Model.ts
@@ -1,7 +1,7 @@
 import _ from "lodash";
-import {cleanOptions, d2BaseModelColumns, d2BaseModelDetails} from "../utils/d2";
-import {TableFilters, TableLabel, TableList, TablePagination} from "../types/d2-ui-components";
-import {D2, ModelDefinition} from "../types/d2";
+import { cleanOptions, d2BaseModelColumns, d2BaseModelDetails } from "../utils/d2";
+import { TableFilters, TableLabel, TableList, TablePagination } from "../types/d2-ui-components";
+import { D2, ModelDefinition } from "../types/d2";
 
 abstract class D2Model {
     // Metadata Type should be defined on subclasses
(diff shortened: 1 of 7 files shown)


SferaDev:
@tokland Please focus on: https://github.com/EyeSeeTea/metadata-synchronization/pull/25/commits/6ec572611fd19b7b6970408285a318f1f2c3d00a

The first commit is a side-effect of the inclusion of TS into the prettify.

src/models/d2Model.ts
```diff
@@ -63,3 +63,9 @@ export class IndicatorModel extends D2Model {
 export class ValidationRuleModel extends D2Model {
     protected static metadataType = "validationRule";
 }
+
+export function defaultModel(pascalCaseModelName: string): any {
```
tokland: This should type: `any`-> `D2Model`.

src/models/d2ModelFactory.js
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * D2ModelProxy allows to create on-demand d2Model classes
+ * If the class doesn't exist a new default class is created
+ * d2ModelName: string (singular name property from d2.models)
+ */
+export function d2ModelFactory(d2ModelName) {
+    const modelName = d2ModelName.charAt(0).toUpperCase() + d2ModelName.slice(1) + "Model";
+    return classes[modelName] ? classes[modelName] : defaultModel(d2ModelName);
+}
```
tokland: TS shines when using those model classes, I am a bit reluctant to leave it as JS, callers won't take advantage of typings/autocomplete. If we could have fixed classes, this should work:

```
type Classes = typeof classes;

export function d2ModelFactory<K extends keyof Classes>(modelName: K): Classes[K] {
    return classes[modelName];
}
```

If having the ability to create classes on-the-fly (as you have know) is a must, this is not so type-restrictive, but  should also work:

```
const classes: { [modelName: string]: D2Model } = {
    OrganisationUnitModel,
    DataElementModel,
    IndicatorModel,
    ValidationRuleModel,
};

export function d2ModelFactory(d2ModelName: string): D2Model {
    const modelName = d2ModelName.charAt(0).toUpperCase() + d2ModelName.slice(1) + "Model";
    return classes[modelName] ? classes[modelName] : defaultModel(d2ModelName);
}
```

SferaDev: Thanks!

SferaDev: Actually we do not return a D2Model instance but the class of D2Model.

```js
function d2ModelFactory(d2ModelName: string): typeof D2Model
```

PR #26, opened 2019-03-14
Several cosmetic changes

matiasgarcia91:
### :pushpin: References

* **Issue:** Closes several issues of #7 

### :memo: Implementation

-

### :art: Screenshots
image

I was going to add a help text but since I'm not completely familiar with the use-case of this app it would probably need changing. We need to define it and I'll update it.

### :bookmark_tabs: Others

- Both in this app and in vaccination we are having a bit of an issue with the spacing on the top of the objectsTable, is it worth it to modify it in the d2-ui library?

### DIFF
--- a/src/components/instance-configurator/InstanceConfigurator.jsx
+++ b/src/components/instance-configurator/InstanceConfigurator.jsx
@@ -3,9 +3,14 @@ import PropTypes from "prop-types";
 import i18n from "@dhis2/d2-i18n";
 import { ObjectsTable, withSnackbar } from "d2-ui-components";
 import { withRouter } from "react-router-dom";
+import { withStyles } from "@material-ui/core/styles";
 import PageHeader from "../shared/PageHeader";
 
 import Instance from "../../models/instance";
+
+const styles = () => ({
+    tableContainer: { marginTop: -10 },
+});
 
 class InstanceConfigurator extends React.Component {
     state = {
@@ -85,12 +90,12 @@ class InstanceConfigurator extends React.Component {
     };
 
     render() {
-        const { d2 } = this.props;
+        const { d2, classes } = this.props;
 
         return (
             <React.Fragment>
                 <PageHeader title={i18n.t("Instances")} onBackClick={this.backHome} />
-                <div style={{ marginTop: -10 }}>
+                <div className={classes.tableContainer}>
                     <ObjectsTable
                         key={this.state.tableKey}
                         d2={d2}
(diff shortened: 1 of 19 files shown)


src/components/instance-form-builder/InstanceFormBuilder.jsx
```diff
@@ -68,7 +68,11 @@ class InstanceFormBuilder extends React.Component {
                     contents={i18n.t("All your changes will be lost. Are you sure?")}
                 />
 
-                <FormHeading title={title} onBackClick={this.cancelSave} />
+                <PageHeader
+                    title={title}
+                    onBackClick={this.cancelSave}
+                    helpText={i18n.t("Help text")}
```
SferaDev: If we don't have a proper help text now we could just remove it for now and define it later on. @adrianq Might have more insight about this.

matiasgarcia91:
@SferaDev thanks for the details! You worked on the ObjectsTable in d2-ui-components, what do you think about this?

> Both in this app and in vaccination we are having a bit of an issue with the spacing on the top of the objectsTable, is it worth it to modify it in the d2-ui library?

src/components/instance-form-builder/InstanceFormBuilder.jsx
matiasgarcia91: We'll wait a bit on this, if not I'll try and write up something coherent

src/components/instance-configurator/InstanceConfigurator.jsx
```diff
@@ -79,24 +80,31 @@ class InstanceConfigurator extends React.Component {
...
-                    pageSize={10}
-                    initialSorting={this.initialSorting}
-                    actions={this.actions}
-                    onCreate={this.onCreate}
-                    list={Instance.list}
-                />
-            </div>
+            <React.Fragment>
+                <PageHeader title={i18n.t("Instances")} onBackClick={this.backHome} />
+                <div style={{ marginTop: -10 }}>
```
tokland: As @SferaDev already noted, let's use `withStyle`

matiasgarcia91: I left this on hold while waiting for the ObjectsTable comment to be answered, I'll change it and we'll remove it when we update d2-ui-components.

src/components/instance-form-builder/InstanceFormBuilder.jsx
adrianq: This is fine for now. We can always revisit it at a later stage. I just want to show the client what we can potentially have and if it makes sense for them

PR #27, opened 2019-03-14
Validate URL-Username uniqueness on instance creation

matiasgarcia91:
### :pushpin: References

* **Issue:** Validate URL-Username uniqueness on instance creation, issue in #7 

### :art: Screenshots
image


SferaDev:
We could change current id => name and add the url-username persistent as a unique field in the dataStore model. What do you think @matiasgarcia91?

src/components/instance-form-builder/GeneralInfoForm.jsx
```diff
@@ -166,12 +166,21 @@ class GeneralInfoForm extends React.Component {
                 return;
             }
 
-            this.setState({ isSaving: true });
+            this.props.instance.validateInstanceName(this.props.d2).then(isValid => {
```
tokland: We tend to favor using async/await for this kind of code to avoid indentations.

src/components/instance-form-builder/GeneralInfoForm.jsx
```diff
@@ -166,12 +166,21 @@ class GeneralInfoForm extends React.Component {
...
+                if (!isValid.status) {
```
tokland: Here we have a unique validation and can be kept as it is. As a rule of thumb, it's preferable to have a single  `model.validate()` that returns something like: `[[FIELD, [msg1, msg2]], [FIELD2, [msg3, msg4]]`. This way the component has less logic.

src/components/instance-form-builder/GeneralInfoForm.jsx
```diff
@@ -166,12 +166,21 @@ class GeneralInfoForm extends React.Component {
...
+                }
 
-            this.props.originalInstance.remove(this.props.d2).then(() => {
-                this.props.instance.save(this.props.d2, this.props.isEdit).then(() => {
-                    this.setState({ isSaving: false });
-                    this.props.history.push("/instance-configurator");
+                this.setState({ isSaving: true });
+
+                this.props.originalInstance.remove(this.props.d2).then(() => {
+                    this.props.instance.save(this.props.d2, this.props.isEdit).then(() => {
```
tokland: It looks like save had a second argument, but now it has only D2 as first argument.

src/models/dataStore.ts
```diff
@@ -1,6 +1,8 @@
 import _ from "lodash";
-import {D2, Response} from "../types/d2";
-import {TableFilters, TableList, TablePagination} from "../types/d2-ui-components";
+import { D2, Response } from "../types/d2";
+import { TableFilters, TableList, TablePagination } from "../types/d2-ui-components";
+import { Data as InstanceData } from "./instance";
+import { SSL_OP_DONT_INSERT_EMPTY_FRAGMENTS } from "constants";
```
tokland: Artifact import

src/models/dataStore.ts
```diff
@@ -82,3 +87,27 @@ export async function deleteInstance(d2: D2, instance: any): Promise<Response> {
...
+): Promise<Response> {
+    try {
+        const instanceArray = await getDataStore(d2, instancesKey);
+        const currentInstanceName = [url, username].join("-");
+        const validCombination = _(instanceArray)
+            .filter(
+                (inst: InstanceData) => [inst.url, inst.username].join("-") === currentInstanceName
+            )
+            .isEmpty()
+            .valueOf();
```
tokland: Can be simplified: `const validCombination = !(instanceArray).some(...)`

src/components/instance-form-builder/GeneralInfoForm.jsx
```diff
@@ -166,12 +166,21 @@ class GeneralInfoForm extends React.Component {
...
+                        this.setState({ isSaving: false });
+                        this.props.history.push("/instance-configurator");
```
tokland: Mental note, also for me, I didn't follow it in vaccination-app: those more-or-less generic components should not have hardcoded urls, it makes more sense to accept a prop `onSave` and let the parent decide what to do. No need to change it here.

src/models/instance.ts
```diff
@@ -42,42 +46,47 @@ export default class Instance {
...
+        return new Instance({ ...this.data, description });
     }
 
     public get description(): string {
         return this.data.description ? this.data.description : "";
     }
+
+    public async validateInstanceName(d2: D2): Promise<Response> {
+        const { url, username } = this.data;
+        return validateInstanceName(d2, url, username);
```
tokland: Problem: this works on a create, but on an update, it will say the combination already exists, when in fact it's the same model we are trying to update.

matiasgarcia91: just a question, should Edit permit all fields to be altered?

matiasgarcia91:
Applied @SferaDev comments on switching to a generated Id for the check but will probably undo after chatting with @tokland about the editing side of the form and how that id would be also changing. Id will be generated with the generateUID() function and validation will revert to field joining

adrianq:
@matiasgarcia91 Remember also to merge with development and fix the conflicts. I will review the functional aspects once you take care of all the comments. Let me know when you are done!

src/models/instance.ts
tokland: As we discussed, let's add an autogenerated ID field to the instances, it will be much easier to update things. Probably `props.originalInstance` could be then removed.

src/models/instance.ts
```diff
@@ -1,11 +1,14 @@
+///<reference path="../types/d2.d.ts" />
```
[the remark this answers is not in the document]
matiasgarcia91: There is an issue with this I discussed yesterday with @tokland. On one hand, module declarations like the one for the generatedUid() must be in a standalone d2.d.ts file (standalone meaning it only has module declarations). On the other hand, the d2 types file must also be .d in order to automatically detect the type declarations like you said.

So I split them up into two different .d files. But if anyone knows a better alternative to this and also to remove the reference (which is needed currently for that file to detect the d2/ui module) let me know

src/models/d2Model.ts
```diff
@@ -47,17 +49,17 @@ abstract class D2Model {
 }
 
 export class OrganisationUnitModel extends D2Model {
-    protected static metadataType = 'organisationUnit';
+    protected static metadataType = "organisationUnit";
```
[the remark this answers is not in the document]
tokland: The current workflow has a lot of room for improvement. What do you think of adding a pre-push hook to prevent a push when there are uncommitted changes?

https://stackoverflow.com/questions/2831172/how-can-i-prevent-git-push-if-local-modifications-are-detected-including-untrac

src/components/instance-form-builder/GeneralInfoForm.jsx
matiasgarcia91: done!

src/components/instance-form-builder/GeneralInfoForm.jsx
matiasgarcia91: I can remove all of these references but I'll prefer to do that on a separate PR in order to not mix to much stuff here

src/models/d2Model.ts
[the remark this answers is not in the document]
matiasgarcia91: Maybe we can create a ticket to tackle both these types of configs as well as the TS type/module export configs?

src/models/instance.ts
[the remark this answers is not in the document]
matiasgarcia91: Renamed the module export file to 

> modules.d.ts
and the types file back to 

> d2.d.ts

src/components/instance-form-builder/GeneralInfoForm.jsx
```diff
@@ -172,21 +171,26 @@ class GeneralInfoForm extends React.Component {
...
+            const fieldKeys = fields.map(field => field.name);
+            const errorMessages = await getValidationMessages(d2, instance, fieldKeys);
 
-            this.setState({ isSaving: true });
-
-            this.props.originalInstance.remove(this.props.d2).then(() => {
-                this.props.instance.save(this.props.d2, this.props.isEdit).then(() => {
-                    this.setState({ isSaving: false });
-                    this.props.history.push("/instance-configurator");
+            if (!_(errorMessages).isEmpty()) {
```
SferaDev: Chaining 1 method can be simplified: _.isEmpty(errorMessages)

src/models/dataStore.ts
```diff
@@ -28,13 +29,16 @@ async function saveDataStore(d2: D2, dataStoreKey: string, newValue: any): Promi
     await dataStore.set(dataStoreKey, newValue);
 }
 
+export async function getDataStoreData(d2: D2, dataStoreKey: string) {
```
SferaDev: I guess this will be part of the dataStore abstraction you mentioned @matiasgarcia91 ?

src/models/instance.ts
```diff
@@ -1,11 +1,16 @@
+///<reference path="../types/modules.d.ts" />
```
SferaDev: Triple-slash directive should not be needed now, tsconfig already looks up for .d.ts files on src and it should correctly find the d2/uid module definition without forcing the compiler to look it up. At least I believe it worked fine when I tested it on my end, could you double check?

src/components/instance-form-builder/GeneralInfoForm.jsx
tokland: I am ok with both ways to write this kind of short lodash code. 

I think that it's acceptable -especially when using a non-chainable lodash method like `_.isEmpty`- because the OOP style is more declarative: `_(myList).isEmpty()`.

SferaDev: Okay then! Will try to have it in mind to keep the same coding style

tokland: Yep, but feel free to use `_.someMethod(var)` whenever it makes sense to you, IMO it's no problem to have those minor nuances in the code :)

src/models/dataStore.ts
matiasgarcia91: yes, I needed this first step so I wrote it here

src/models/instance.ts
matiasgarcia91: sure

src/models/dataStore.ts
```diff
@@ -76,7 +79,11 @@ export async function saveNewInstance(d2: D2, instance: any): Promise<Response>
 export async function deleteInstance(d2: D2, instance: any): Promise<Response> {
     try {
         const instanceArray = await getDataStore(d2, instancesKey);
-        const newInstanceArray = _.differenceWith(instanceArray, [instance], _.isEqual);
+        const newInstanceArray = _.differenceWith(
+            instanceArray,
+            [instance],
+            (inst: InstanceData) => inst.id === instance.id
+        );
```
tokland: Proposal:

```
        const instanceArray: InstanceData[] = await getDataStore(d2, instancesKey);
        const newInstanceArray = instanceArray.filter(inst => inst.id !== instance.id);
```

matiasgarcia91: I'll change this logic in PR #31 where the datastore gets altered

matiasgarcia91: done

src/models/d2Model.ts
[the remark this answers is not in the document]
adrianq: @matiasgarcia91 did you finally create a ticket for this?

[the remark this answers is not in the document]
matiasgarcia91: no, TS configs seem to be solved but I'll create one to unify configs between both apps, what do you think?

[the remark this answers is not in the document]
adrianq: Yep, go ahead!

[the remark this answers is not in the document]
matiasgarcia91: https://github.com/EyeSeeTea/metadata-synchronization/issues/38

PR #30, opened 2019-03-20
Add confirmation dialog to delete instance action

matiasgarcia91:
### :art: Screenshots
image

This PR is based against PR #27

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2019-03-22T11:16:32.473Z\n"
-"PO-Revision-Date: 2019-03-22T11:16:32.473Z\n"
+"POT-Creation-Date: 2019-03-22T11:28:52.570Z\n"
+"PO-Revision-Date: 2019-03-22T11:28:52.570Z\n"
 
 msgid "Metadata Synchronization"
 msgstr ""
@@ -38,10 +38,10 @@ msgstr ""
 msgid "Ok"
 msgstr ""
 
-msgid "Deleted"
+msgid "Deleted {{name}}"
 msgstr ""
 
-msgid "Failed to delete"
+msgid "Failed to delete {{name}}"
 msgstr ""
 
 msgid "Server name"
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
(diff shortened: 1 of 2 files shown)


i18n/en.pot
```diff
@@ -59,6 +59,12 @@ msgstr ""
 msgid "Delete"
 msgstr ""
 
+msgid "Delete Instance?"
+msgstr ""
+
+msgid "Are you sure you want to delete this instance?"
+msgstr ""
+
```
tokland: Are those two literals missing for es.po?

src/components/instance-configurator/InstanceConfigurator.jsx
```diff
@@ -15,6 +17,8 @@ const styles = () => ({
 class InstanceConfigurator extends React.Component {
     state = {
         tableKey: Math.random(),
+        dialogOpen: false,
+        toDelete: {},
```
tokland: It's usually more declarative to use `null` instead of an empty object (and easier to work with the var, also)

src/components/instance-configurator/InstanceConfigurator.jsx
```diff
@@ -15,6 +17,8 @@ const styles = () => ({
 class InstanceConfigurator extends React.Component {
     state = {
         tableKey: Math.random(),
+        dialogOpen: false,
```
tokland: Proposal: Remove `dialogOpen` and use `toDelete` to determine if the dialog should be open or not.

matiasgarcia91: like it

i18n/en.pot
matiasgarcia91: there is no es.po file in this project yet, I'll generate one on a separate PR

adrianq: No need to generate any es.po. They don't want multilingual as this app is planned to be used by English speakers

src/components/instance-configurator/InstanceConfigurator.jsx
```diff
@@ -34,17 +37,27 @@ class InstanceConfigurator extends React.Component {
...
+    };
+
+    handleDialogConfirm = async () => {
+        const { toDelete } = this.state;
+        const instance = new Instance(toDelete);
+        this.setState({ toDelete: null });
         await instance.remove(this.props.d2).then(response => {
             if (response.status) {
-                this.props.snackbar.success("Deleted " + instanceData.name);
+                this.props.snackbar.success("Deleted " + toDelete.name);
```
adrianq: Why don't we add this to i18n?

src/components/instance-configurator/InstanceConfigurator.jsx
```diff
@@ -34,17 +37,27 @@ class InstanceConfigurator extends React.Component {
...
 
                 // TODO: Add a way to force render of ObjectsTable on-demand
                 // This is a work-around
                 this.setState({ tableKey: Math.random() });
             } else {
-                this.props.snackbar.error("Failed to delete " + instanceData.id);
+                this.props.snackbar.error("Failed to delete " + toDelete.name);
```
adrianq: and this?

matiasgarcia91: sure, my bad!

PR #32, opened 2019-03-20
Add instance connection test to save

matiasgarcia91:
### :memo: Implementation -
Thanks to @tokland for the code snippet!

### :art: 
image

This PR is based against PR #27

### DIFF
--- a/src/components/instance-form-builder/GeneralInfoForm.jsx
+++ b/src/components/instance-form-builder/GeneralInfoForm.jsx
@@ -180,9 +180,13 @@ class GeneralInfoForm extends React.Component {
             }
             const fieldKeys = fields.map(field => field.name);
             const errorMessages = await getValidationMessages(d2, instance, fieldKeys);
+            const connectionErrors = await instance.check();
 
-            if (!_(errorMessages).isEmpty()) {
-                this.props.snackbar.error(errorMessages.join("\n"), {
+            const allErrors = connectionErrors.status
+                ? errorMessages
+                : [...errorMessages, connectionErrors.error];
+            if (!_(allErrors).isEmpty()) {
+                this.props.snackbar.error(allErrors.join("\n"), {
                     autoHideDuration: null,
                 });
             } else {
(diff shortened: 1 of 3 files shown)


src/components/instance-form-builder/GeneralInfoForm.jsx
```diff
@@ -180,9 +180,13 @@ class GeneralInfoForm extends React.Component {
             }
             const fieldKeys = fields.map(field => field.name);
             const errorMessages = await getValidationMessages(d2, instance, fieldKeys);
+            const connectionErrors = await instance.check();
```
matiasgarcia91: Do you think this should be all part of the validations flow or separate like it's here?

src/models/instance.ts
```diff
@@ -155,4 +155,39 @@ export default class Instance {
...
     }
+
+    public async check(): Promise<Response> {
+        const { url, username, password } = this.data;
+        const headers = new Headers({
+            Authorization: "Basic " + btoa(username + ":" + password),
+        });
+
+        try {
+            const res = await fetch(url + "/api/system/info", { method: "GET", headers });
```
SferaDev: If we keep it in the models it should be window agnostic to be reused on Phase 2.

Something like this would also work without fetch nor btoa which are not available on node. Note that setting the headers is not needed because axios does it alone but beware that the response object will change a little. 

```js
axios.get(url + "/api/system/info", { auth: {
    username, 
    password, 
  }});
```

Detecting the CORS issue is not possible in axios either, I tried and not succeed so the if conditions you have are good enough.

src/components/instance-form-builder/GeneralInfoForm.jsx
SferaDev: I guess this should be moved to save() and handle this kind of validations there.

We could find potential performance issues if the user does loads of changes and not being able to cancel running Promises.

tokland: There are some valid alternatives. As Alexis says, it makes sense to move the logic the model. I'd rename it to `checkAndSave()` (calls check + save), and keep a `save()` without the check.

Another functional option is to add a separate button that just tests the instance and leave the save without a check. Use case: you want to create an instance record, but the instance is not up at the moment for some reason. Check this with @adrianq, though.

src/models/instance.ts
tokland: There are some isomorphic (node/browser) fetch packages, but since we are already using axios in this app, I'd go with axios.

PR #45, opened 2019-03-26
Add level to OU ObjectsTable

### LINKED ISSUE #43: Add Level to OU Object Table

SferaDev:
### :pushpin: References

* **Issue:** Closes #43 

### :memo: Implementation

- Add custom columns/details for OU

- The abstract model handles the logic behind to add the columns, order, list...

### :art: Screenshots

image

image

- Branch based on top of #37 to minimize conflicts. Should be merged afterwards.

### DIFF
--- a/src/utils/d2.ts
+++ b/src/utils/d2.ts
@@ -26,8 +26,15 @@ export const d2BaseModelDetails = [
 ];
 
 export const organisationUnitsDetails = [
-    { name: "level", text: i18n.t("Level"), sortable: true },
-    ...d2BaseModelDetails,
+    { name: "displayName", text: i18n.t("Name") },
+    { name: "shortName", text: i18n.t("Short name") },
+    { name: "code", text: i18n.t("Code") },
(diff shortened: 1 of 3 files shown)


src/utils/d2.ts
```diff
@@ -19,6 +25,11 @@ export const d2BaseModelDetails = [
     { name: "href", text: i18n.t("API link") },
 ];
 
+export const organisationUnitsDetails = [
+    { name: "level", text: i18n.t("Level"), sortable: true },
+    ...d2BaseModelDetails,
```
tokland: Not sure, do we mind that *Level* is the first item instead of the traditional *Name*?

SferaDev: Maintenance doesn't even add the level to details, just to columns.

@adrianq Would you remove it from details too?

SferaDev: As discussed with Adrian, we will keep it and lower it down.

PR #46, opened 2019-03-26
Encrypt password when saving to dataStore

### LINKED ISSUE #7: Implement instances configuration

matiasgarcia91:
* **Issue:** Closes #7

### :memo: Implementation-
Encrypts passwords before saving and decrypts it when loading the edit page. 

The README has updated, now the file `public/app-config.json` must be created based on the `public/app-config.template.json` file.
Also, careful with pre-existing instances which do not have their passwords encrypted as they break the app when trying to edit them. Best to remove them to avoid future issues once this is merged


README.md
```diff
@@ -4,6 +4,8 @@
 $ yarn install
 ```
 
+File public/app-config.json must be created by duplicating public/app-config.template.json and filling in the encryptionKey.
```
tokland: [minor] Maybe some backticks on files and keys to highlight them: `public/app-config.js`, and so on

src/models/instance.ts
```diff
@@ -103,6 +103,18 @@ export default class Instance {
...
 
+    public encryptPassword(encryptionKey: string) {
+        const cryptr = new Cryptr(encryptionKey);
+        const encrypted = cryptr.encrypt(this.data.password);
+        return new Instance({ ...this.data, password: encrypted });
+    }
+
+    public decryptPassword(encryptionKey: string) {
+        const cryptr = new Cryptr(encryptionKey);
+        const decrypted = cryptr.decrypt(this.data.password);
```
tokland: When I try to edit an existing instance I get: `TypeError: invalid iv length 1`. That's normal, it tried to decrypt a plain password. Can you add a note in the section "notes for the tester" that old instances cannot be edited (should be deleted, I guess)?

cypress/integration/landing-page.spec.js
```diff
@@ -2,6 +2,8 @@
 
 context("Landing page", () => {
     before(() => {
+        cy.server();
+        cy.fixture("app-config.json").then(json => cy.route("GET", "app-config.json", json));
```
tokland: Cool!

src/types/modules.d.ts
```diff
@@ -2,6 +2,14 @@ declare module "d2/uid" {
     export function generateUid(): string;
 }
 
+declare module "cryptr" {
```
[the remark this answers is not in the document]
matiasgarcia91: cool! I was kinda looking for this for a while

src/models/instance.ts
matiasgarcia91: sure!

src/models/instance.ts
```diff
@@ -34,6 +35,17 @@ export default class Instance {
         return new Instance(initialData);
     }
 
+    public static getOrCreate(data: Data, encryptionKey: string): Instance {
```
tokland: `data` may be undefined: `getOrCreate(data: Data | undefined, encryptionKey: string`. In vaccination we use the type `Maybe` (usual in many languages): `type Maybe<T> = T | undefined;`. Both ways are ok.

src/models/instance.ts
```diff
@@ -34,6 +35,17 @@ export default class Instance {
...
+        let instance;
+        if (!!data) {
```
tokland: While this is conceptually perfect, `if (boolean)`, in dynamic languages is ok to leave the object: `if (object)`, because we simply are checking if it has a "truthy" value (according to the language semantics). It's also visually less distracting.

PR #47, opened 2019-03-26
Add separate button to test connection as well as contextual action

### LINKED ISSUE #33: Improve connection validation
it makes sense to move the logic the model. I'd rename it to checkAndSave() (calls check + save), and keep a save() without the check.

Another functional option is to add a separate button that just tests the instance and leave the save without a check. Use case: you want to create an instance record, but the instance is not up at the moment for some reason.

matiasgarcia91:
### :pushpin: References
* **Issue:** Closes #33 

### :memo: Implementation
Connection validation is no longer performed on save. There is now a separate button on the form to validate the connection as well as a new contextual action on the instancesList. 

### :art: Screenshots
image
image
image

This branch is based off the `encrypt-password-7` in order to integrate encryption into the connection validation. Same warnings as that branch, `app-config.json` file must be created and older instances created before this PR will break the app when trying to `Edit` or `Test Connection` from the contextual options

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2019-03-22T11:28:52.570Z\n"
-"PO-Revision-Date: 2019-03-22T11:28:52.570Z\n"
+"POT-Creation-Date: 2019-03-26T16:05:15.265Z\n"
+"PO-Revision-Date: 2019-03-26T16:05:15.265Z\n"
 
 msgid "Metadata Synchronization"
 msgstr ""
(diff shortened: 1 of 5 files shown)


i18n/en.pot
```diff
@@ -38,6 +38,9 @@ msgstr ""
 msgid "Ok"
 msgstr ""
 
+msgid "Connected sucessfully to instance"
```
tokland: Typo: successfully

src/components/instance-form-builder/GeneralInfoForm.jsx
```diff
@@ -199,6 +199,25 @@ class GeneralInfoForm extends React.Component {
             }
         };
 
+        const testConnection = async () => {
```
tokland: Move it to a class method to lighten the render method and to keep the prop constant.

matiasgarcia91: This was my first approach but I lacked the `fields` to perform the form validation before checking the connection (to make sure url, username, pass, were entered). I could move the bulk of the method to a class method and just call the form validation and then the class method from the render

PR #52, opened 2019-04-01
Add instance id to edit instance url, reload instance on page refresh

### LINKED ISSUE #29: Edit page for instances
The edit URL should include the identifier of the element being edited. 

Check also that this new URL works on a page load.

matiasgarcia91:
### :pushpin: References

* **Issue:** Closes #29

### :memo: Implementation
Included instance id on url and perform the instance get on page refresh

### :art: Screenshots
image

### DIFF
--- a/src/components/instance-form-builder/InstanceFormBuilder.jsx
+++ b/src/components/instance-form-builder/InstanceFormBuilder.jsx
@@ -17,21 +17,20 @@ class InstanceFormBuilder extends React.Component {
         appConfig: PropTypes.object.isRequired,
     };
 
+    isEdit = this.props.match.params.formFor === "edit";
+
     constructor(props) {
         super(props);
 
         const {
-            match,
             location,
             appConfig: { encryptionKey },
         } = props;
 
-        const isEdit = match.params.formFor === "edit";
         const instance = Instance.getOrCreate(location.instance, encryptionKey);
 
         this.state = {
             instance,
-            isEdit,
             dialogOpen: false,
         };
     }
@@ -42,8 +41,8 @@ class InstanceFormBuilder extends React.Component {
             appConfig: { encryptionKey },
             d2,
         } = this.props;
-        const { instance, isEdit } = this.state;
-        if (isEdit && !instance.data.id) {
+        const { instance } = this.state;
+        if (this.isEdit && !instance.data.id) {
             const encryptedInstance = await Instance.get(d2, params.id);
             const instanceToEdit = encryptedInstance.decryptPassword(encryptionKey);
             this.setState({ instance: instanceToEdit });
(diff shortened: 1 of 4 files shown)


src/components/instance-form-builder/InstanceFormBuilder.jsx
```diff
@@ -21,11 +21,12 @@ class InstanceFormBuilder extends React.Component {
         super(props);
 
         const {
+            match,
             location,
             appConfig: { encryptionKey },
         } = props;
 
-        const isEdit = location.instance !== undefined;
+        const isEdit = match.params.formFor === "edit";
```
tokland: Property `isEdit` depends solely on `props`, strictly speaking, it should not be in the state, maybe a class constant?

src/components/instance-form-builder/InstanceFormBuilder.jsx
```diff
@@ -35,6 +36,20 @@ class InstanceFormBuilder extends React.Component {
...
     }
 
+    componentDidMount = async () => {
+        const {
+            match: { params },
+            appConfig: { encryptionKey },
+            d2,
+        } = this.props;
+        const { instance, isEdit } = this.state;
+        if (isEdit && !instance.data.id) {
```
tokland: I see we have a `Instance.get` and `Instance.getOrCreate`, one with d2 but no encryption key, the other with encryption key but not d2 (so it does queries the DB). Ideally, we should have a single `async Instance.getOrCreate(d2, id, encryptionKey)`. Then this code would be reduced to a single line.

src/models/instance.ts
```diff
@@ -55,7 +55,8 @@ export default class Instance {
     }
 
     public static async get(d2: D2, id: string): Promise<Instance> {
-        return (await getDataById(d2, instancesDataStoreKey, id)) as Instance;
+        const instance = await getDataById(d2, instancesDataStoreKey, id);
+        return new Instance(instance);
```
tokland: Good catch!

src/components/instance-form-builder/InstanceFormBuilder.jsx
matiasgarcia91: I agree, it's inconsistent this way and a bit tangled up. I'm going to take a look at it

matiasgarcia91: @SferaDev has asked to refactor this so he'll do it in another task

SferaDev:
The whole getOrCreate needs to be refactored and the password encryption broke the instance.check() method.

matiasgarcia91:
instance.check() method fails if you check a connection which was not created with the same encryptionKey as the one you are using now. We should add some error control to prevent the app from crashing

adrianq:
As discussed with @SferaDev we will for now just decript the password before sync the OUs and create a card to refactor the Model and centralise encript/decript process

SferaDev:
> instance.check() method fails if you check a connection which was not created with the same encryptionKey as the one you are using now. We should add some error control to prevent the app from crashing

Okey, that could be. Sounds good as part of the refactor.

matiasgarcia91:
yes, it would be more of a safeguard since this should only happen to us while developing since we all have different keys but it'd be nice to have either way

SferaDev:
> yes, it would be more of a safeguard since this should only happen to us while developing since we all have different keys but it'd be nice to have either way

The app should use default key to avoid this and all development branches should use it instead of overriding. What do you think?

matiasgarcia91:
yeap! that sounds great. An easy way would be to add it to the app-config.template.json but idk if putting even a dev key on GH is okey

adrianq:
For the record, this is the card we created to refactor the encrypt/decrypt https://github.com/EyeSeeTea/metadata-synchronization/issues/54

PR #58, opened 2019-04-05
Disable synchronization if no OU or target instance selected

### LINKED ISSUE #50: Improve validation on OU sync
- [ ] If no OU is selected synchronization should not be available (grey fields, snackbar message or any other option)
- [ ] If no target instance is selected disable Synchronization button

matiasgarcia91:
### :pushpin: References

* **Issue:** Closes #50

### :art: Screenshots
image

The disabling of the button inside the dialog won't work until d2-ui-comp is updated.

### :bookmark_tabs: Others
Yes, added a prop to the confirmationDialog to control disabling of the confirmation button/action.
https://github.com/EyeSeeTea/d2-ui-components/pull/27


src/components/shared/Dropdown.jsx
```diff
@@ -73,5 +73,5 @@ Dropdown.propTypes = {
     items: PropTypes.array.isRequired,
     value: PropTypes.string.isRequired,
     onChange: PropTypes.func.isRequired,
-    placeholder: PropTypes.string.isRequired,
+    label: PropTypes.string.isRequired,
```
SferaDev: This is also on another PR, no problem as merge will take care.

src/components/sync-configurator/BaseSyncConfigurator.jsx
```diff
@@ -152,6 +153,7 @@ class BaseSyncConfigurator extends React.Component {
                         size="large"
                         onClick={this.onSynchronize}
                         data-test="list-action-bar"
+                        disabled={disabled}
```
SferaDev: Wouldn't a hidden make more sense? For the FAB only

src/components/sync-configurator/BaseSyncConfigurator.jsx
```diff
@@ -124,7 +125,7 @@ class BaseSyncConfigurator extends React.Component {
             filters,
             importResponse,
         } = this.state;
-
+        const disabled = metadata.organisationUnit && _.isEmpty(metadata.organisationUnit);
```
tokland: BaseSyncConfigurator is a generic component, we should try not to hardcode model names. Proposal for a more generic approach: 

```const disabled = _(metadata).values().some(_.isEmpty)```

I have some doubts about the type of `metadata`, we are using an object with has always has a single key. You usually would describe it simply with a pair (array) or an object. Example:

We have: ```const metadata = {organisationUnits: []}```
Option 1 (pair): ```const metadata = ["organisationUnits", []]```
Option 2 (object), usually the best: ```const metadata = {model: "organisationUnits", objects: []}```

Not that we have to change it, just a note for future code.

File: src/components/sync-configurator/BaseSyncConfigurator.jsx (hunk 2 above)
Comment (tokland) [reply src/components/sync-configurator/BaseSyncConfigurator.jsx]: Let's recap the options:

1) Keep the button enable and show a snackbar message
2) [what we have here] disable the button
3) Hide the button

I have no strong opinion about this, I'll leave the decision for @adrianq

File: src/components/sync-dialog/SyncDialog.jsx [hunk 4]
```diff
@@ -61,7 +66,8 @@ class SyncDialog extends React.Component {
 
     render() {
         const { d2, isOpen } = this.props;
-
+        const { targetInstances } = this.state;
+        const disableSync = targetInstances && _.isEmpty(targetInstances);
         return (
```
tokland: Minor: let's keep a blank space before the return, it usually keeps function/methods more clear.

src/components/sync-dialog/SyncDialog.jsx
```diff
@@ -61,7 +66,8 @@ class SyncDialog extends React.Component {
 
     render() {
         const { d2, isOpen } = this.props;
-
+        const { targetInstances } = this.state;
+        const disableSync = targetInstances && _.isEmpty(targetInstances);
```
tokland: Minor: AFICS, `targetInstances` is always an array, so the first check can be removed.

src/components/sync-configurator/BaseSyncConfigurator.jsx
SferaDev: The thing is that we are following how the import/export API works. The result of a metadata export is like that { metadataType: [] }. So we have it that way in code to reuse the response of DHIS2 without having to transform it.

src/components/sync-configurator/BaseSyncConfigurator.jsx
matiasgarcia91: I think leaving the greyed out button there is better, at least at first glance you can know there is some action you can perform on that page that is currently disabled.

SferaDev: Maybe with styles we could darken it differently.

But I still prefer adding a snackbar message without disabling completely (UX wise) or a hide/show animation when selection happens.

adrianq: I don't have a strong opinion between disabling or showing a snackbar. One of the problems now is that the icon is not easily visible when disabled and can be a bit confusing for the user. Anyway, I would go for a snackbar and keep the same styling always. We can show a message telling the user to select at least one OU. Can you please take care of this @matiasgarcia91 ?

In case anybody is interested in the (non-existent) consensus https://ux.stackexchange.com/questions/83645/hiding-vs-disabling-actions-and-show-a-message-on-mobile

I will merge this https://github.com/EyeSeeTea/d2-ui-components/pull/27 anyway as it could be useful in the future

PR #67, opened 2019-04-09
Move back to v30 header-bar

SferaDev:
### :art: Screenshots

image

- We're back using default v30 header bar with translations fixed.

- In the future we might migrate to the new [@dhis2/ui-widgets](https://github.com/dhis2/ui-widgets) HeaderBar but the style is v32.


SferaDev:
> Code looks good to me. A visual detail I think we already had: when a user has messages, the badge goes out of the screen:
> 
> Screenshot from 2019-04-10 10-01-52

I'm not sure if we can fix it with styles.

PR #69, opened 2019-04-16
Add filters for Indicators and ValidationRules

### LINKED ISSUE #65: Add Date and Indicator Group Filter to Indicator Sync

### LINKED ISSUE #66: Add Date and Validation Group Filter to Validation Rule Sync

SferaDev:
### :pushpin: References

* **Issue:** Closes #65 and Closes #66 

### :memo: Implementation

- Abstract group and level filters

- Add filters for Validation Rules

- Add filters for Indicators

- Refactor Organisation Unit filters

### :art: Screenshots

image

image

image

image

image

- Memoization is gone in this PR, I'm not sure where we should place it in the state.

### DIFF
--- a/src/components/sync-configurator/BaseSyncConfigurator.jsx
+++ b/src/components/sync-configurator/BaseSyncConfigurator.jsx
@@ -31,8 +31,6 @@ class BaseSyncConfigurator extends React.Component {
             lastUpdatedDate: null,
             groupFilter: null,
             groupFilterData: [],
-            levelFilter: null,
-            levelFilterData: [],
         },
         metadata: {},
         importResponse: [],
@@ -49,14 +47,12 @@ class BaseSyncConfigurator extends React.Component {
         renderExtraFilters: PropTypes.func,
         extraFiltersState: PropTypes.object,
         groupFilterName: PropTypes.string,
-        levelFilterName: PropTypes.string,
     };
 
     static defaultProps = {
         renderExtraFilters: null,
         extraFiltersState: null,
         groupFilterName: null,
-        levelFilterName: null,
     };
 
     actions = [
@@ -69,13 +65,11 @@ class BaseSyncConfigurator extends React.Component {
     ];
 
     componentDidMount = async () => {
-        const { groupFilterName, levelFilterName } = this.props;
         const { filters } = this.state;
 
-        const groupFilterData = groupFilterName ? await this.getGroupFilterData() : null;
-        const levelFilterData = levelFilterName ? await this.getLevelFilterData() : null;
+        const groupFilterData = await this.getGroupFilterData();
 
-        this.setState({ filters: { ...filters, groupFilterData, levelFilterData } });
+        this.setState({ filters: { ...filters, groupFilterData } });
     };
 
     componentDidUpdate = prevProps => {
@@ -85,11 +79,11 @@ class BaseSyncConfigurator extends React.Component {
         }
     };
 
-    onBackHome = () => {
+    goHome = () => {
         this.props.history.push("/");
     };
 
-    onSelectionChange = metadataSelection => {
+    changeSelection = metadataSelection => {
         const metadata = {
             [this.props.model.getMetadataType()]: metadataSelection,
         };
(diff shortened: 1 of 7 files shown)


src/components/sync-configurator/BaseSyncConfigurator.jsx
```diff
@@ -58,28 +68,61 @@ class BaseSyncConfigurator extends React.Component {
...
+
     componentDidUpdate = prevProps => {
         const { extraFiltersState } = this.props;
         if (extraFiltersState !== prevProps.extraFiltersState) {
             this.setState({ filters: { ...this.state.filters, ...extraFiltersState } });
         }
     };
 
-    backHome = () => {
+    onBackHome = () => {
```
tokland: Note (not a code change request): It's usually more declarative to use a verb for the method instead of replicating the event name/prefix. Then the code may read: `<Link onClick={this.goHome}`.

src/components/sync-configurator/BaseSyncConfigurator.jsx
```diff
@@ -95,30 +138,75 @@ class BaseSyncConfigurator extends React.Component {
...
-        if (importResponse) {
-            this.setState({ syncDialogOpen: false, syncSummaryOpen: true, importResponse });
-        } else {
-            this.props.snackbar.error(i18n.t("Unknown error with the request"));
-            this.setState({ syncDialogOpen: false });
-        }
+    getGroupFilterData = async () => {
+        const { d2, groupFilterName } = this.props;
+
+        const groupClass = d2ModelFactory(d2, groupFilterName);
```
tokland: `props.groupFilterName` is optional. I know the condition has been tested by the caller, but a method should work on isolation. So one option is checking here the prop instead of the parent, and return a null (so the parent is simplified). Idem for getLevelFilterData.

src/components/sync-configurator/BaseSyncConfigurator.jsx
```diff
@@ -153,7 +242,7 @@ class BaseSyncConfigurator extends React.Component {
                         list={list}
                         customFiltersComponent={this.renderCustomFilters}
                         customFilters={filters}
-                        onSelectionChange={this.selectionChange}
```
tokland: Following the previous idea, that would be: `onSelectionChange={this.changeSelection}`. Does it make sense?

src/models/d2Model.ts
```diff
@@ -88,6 +92,9 @@ export abstract class D2Model {
 
 export class OrganisationUnitModel extends D2Model {
     protected static metadataType = "organisationUnit";
+    protected static groupFilterName = "organisationUnitGroups";
+    protected static levelFilterName = "level";
```
tokland: Just a comment, if you are convinced it's better (or simpler) this way, go ahead: isn't a bit odd to have references in the abstract class to filter fields not present in all models? (groups, levels, more that we could potentially have).

The most simple alternative I see: a method on the derived classes that is called (if defined) from listMethod with the current object filter and returns an array of d2 filters.

src/components/sync-configurator/BaseSyncConfigurator.jsx
SferaDev: Looks good suggestion for me, in this case we had without verb, handle and on. To make it all consistent I renamed all the methods to use the on but makes sense what you point.

src/models/d2Model.ts
SferaDev: I thought of it too, groups makes sense since DHIS2 has multiple metadata grouped and all of them uses the same approach. Here we're re-using the same logic for Indicators, ValidationRules and OrganisationUnits but if we expanded the reusability would be even more.

With levels I agree that it should probable be moved to list custom filters, there're only 2 metadata classes that allow level filtering: OrganisationUnits and DataApproval.

Even though I left it inside the generic because both of them use the same filtering approach but I should probable move it outside.

src/components/sync-configurator/BaseSyncConfigurator.jsx
SferaDev: Indeed

SferaDev:
I'll fix the comments during the morning and re-request review when done. Thanks @tokland

PR #72, opened 2019-04-25
Add Integration tests for Instance creation and metadata synchronization & Update travis config to actually wait for server to load

SferaDev:
### :memo: Implementation

- Wait-on module has a false-positive with webpack-dev-server. Instead of waiting for http, we now wait for http-get. See https://github.com/bahmutov/start-server-and-test/blob/master/README.md#note-for-webpack-dev-server-users

- REACT_APP_CYPRESS=true was not set during server boot.

- We had a problem that randomly travis crashed because 10s was not enough for server to boot, this is now fixed with the wait-on.

- Though this won't fix the issues we're having with cypress yet but it will likely make the travis testing more robust.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ install:
 script:
   - PORT=8081 REACT_APP_DHIS2_BASE_URL=http://dev2.eyeseetea.com:8083 REACT_APP_CYPRESS=true yarn start &
   - yarn wait-on http-get://localhost:8081
-  - CYPRESS_EXTERNAL_API=http://dev2.eyeseetea.com:8083 CYPRESS_ROOT_URL=http://localhost:8081 yarn cy:e2e:run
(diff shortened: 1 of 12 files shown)


.travis.yml
```diff
@@ -1,18 +1,18 @@
...
 install:
-- yarn install --frozen-lockfile
-- yarn cy:verify
-- yarn build
+  - yarn install --frozen-lockfile
+  - yarn cy:verify
+  - yarn build
 script:
-    - PORT=8081 REACT_APP_DHIS2_BASE_URL=http://dev2.eyeseetea.com:8083 yarn start &
-    - yarn run wait-on http://localhost:8081 && sleep 10s
```
tokland: So I guess travis add node_modules binaries to tthe path and we don't need `yarn run xyz` / `npx xyz`.

PR #75, opened 2019-05-07
Store and Visualize Synchronization Notifications

### LINKED ISSUE #12: Implement notification/history page
# Type definitions

- [x] Create new common SynchronizationResults type:

<details><summary>Original idea</summary>

```
   - SynchronizationResult: Array
   - Timestamp 
   - User
   - Status (READY, IN PROGRESS, FAILURE AND DONE)
   - metadata: {
            metadataType1: ["myId1"],
            metadataType2: ["myId2"],
        }
```

</details>

<details><summary>Implementation</summary>

```ts
export interface SynchronizationResults {
    results: SynchronizationResult[];
    timestamp: Date;
    user: string;
    status: "READY" | "RUNNING" | "FAILURE" | "DONE";
    metadata: MetadataPackage;
}

export interface MetadataPackage {
    [metadataType: string]: any[];
}
```

</details>

- [x] Update existing SynchronizationResult type:

<details><summary>Original idea</summary>

```
   - instance: {id: string; name: string; url: string; **user**: string };
   - status: string; (PENDING, OK, WARNING, ERROR, NETWORK ERROR)
   - importParams: MetadataImportParams;
   - typeReports: any[];
   - stats: {created: number; deleted: number; ignored: number; total: number; updated: number;};
   - report: { typeStats: any[]; messages: any[]; };
```

</details>

<details><summary>Implementation</summary>

```ts
export interface SynchronizationResult extends MetadataImportResponse {
    instance: {
        id: string;
        name: string;
        url: string;
    };
    report: {
        typeStats: MetadataImportStats[];
        messages: any[];
    };
}

export interface MetadataImportResponse {
    status: "PENDING" | "OK" | "WARNING" | "ERROR" | "NETWORK ERROR";
    importParams: MetadataImportParams;
    typeReports: any[];
    stats: MetadataImportStats;
}

export interface MetadataImportStats {
    created: number;
    deleted: number;
    ignored: number;
    updated: number;
    total: number;
}
```
</details>

# Workflow

- [x] 1) Insert SynchronizationResultSet 'in progress' and SynchronizationResult with instance and status 'pending'.

- [x] 2) When each sync process finished we update SynchronizationResult to DHIS2 response status and update the rest of the fields.

- [x] 3) When all sync processes are done we update the status of the SynchronizationResultSet.

# GUI (Object Table)

   - Filter: Text Search Filter, Date Filter and Status Filter
   - Columns: User, Timestamp, Status
   - Contextual action:
      - Details: User, Timestamp, Status, Metadata (Types and Ids), URL to Metadata end point?
      - View Summary: Reuse sync summary and display in a modal.
      - Delete: Delete it from data store

SferaDev:
### :pushpin: References

* **Issue:** Closes #12 

### :memo: Implementation

- Create new Notifications Page and Route

- Streamline Instance Configuration with the updates done for Notifications

- Create a new SyncReport that can be stored in dataStore

- Make SyncSummary more robust to not crash with failed imports

- Update synchronization logic to insert syncReports during the process

- Misc fixes for edge-cases on axios calls

### :art: Screenshots

image

image

image

image

### :bookmark_tabs: Others

- We have a bug in d2-ui-components that when getValue is specified on the details box, the formatter is ignored. Since I want to discuss the approach in the next weekly meeting I introduced a workaround to have everything working until we decide how to fix it properly.

### DIFF
--- a/src/components/notifications-table/NotificationsTable.jsx
+++ b/src/components/notifications-table/NotificationsTable.jsx
@@ -10,7 +10,6 @@ import PageHeader from "../shared/PageHeader";
 
 import SyncReport from "../../models/syncReport";
 import SyncSummary from "../sync-summary/SyncSummary";
-import { createMetadataExportUrl } from "../../utils/synchronization";
 
 const styles = () => ({
     tableContainer: { marginTop: -10 },
--- a/src/components/notifications-table/NotificationsTable.jsx
+++ b/src/components/notifications-table/NotificationsTable.jsx
@@ -177,14 +177,6 @@ class NotificationsTable extends React.Component {
 
         return (
             <React.Fragment>
-                <ConfirmationDialog
-                    isOpen={!!toDelete}
-                    onSave={this.confirmDelete}
-                    onCancel={this.cancelDelete}
-                    title={i18n.t("Delete Instance?")}
-                    description={i18n.t("Are you sure you want to delete this instance?")}
-                    saveText={i18n.t("Ok")}
-                />
                 <PageHeader title={i18n.t("Notifications")} onBackClick={this.backHome} />
                 <div className={classes.tableContainer}>
                     <ObjectsTable
@@ -199,10 +191,20 @@ class NotificationsTable extends React.Component {
                         list={SyncReport.list}
                     />
                 </div>
+
                 <SyncSummary
                     response={syncReport}
                     isOpen={summaryOpen}
                     handleClose={this.closeSummary}
+                />
+
+                <ConfirmationDialog
+                    isOpen={!!toDelete}
+                    onSave={this.confirmDelete}
+                    onCancel={this.cancelDelete}
+                    title={i18n.t("Delete Instance?")}
+                    description={i18n.t("Are you sure you want to delete this instance?")}
+                    saveText={i18n.t("Ok")}
                 />
             </React.Fragment>
         );
--- a/src/components/notifications-table/NotificationsTable.jsx
+++ b/src/components/notifications-table/NotificationsTable.jsx
@@ -1,5 +1,6 @@
 import React from "react";
 import PropTypes from "prop-types";
+import _ from "lodash";
 import i18n from "@dhis2/d2-i18n";
 import { ObjectsTable, withSnackbar, withLoading, ConfirmationDialog } from "d2-ui-components";
 import { withRouter } from "react-router-dom";
@@ -9,6 +10,7 @@ import PageHeader from "../shared/PageHeader";
 
 import SyncReport from "../../models/syncReport";
 import SyncSummary from "../sync-summary/SyncSummary";
+import { createMetadataExportUrl } from "../../utils/synchronization";
 
 const styles = () => ({
     tableContainer: { marginTop: -10 },
@@ -27,6 +29,14 @@ class NotificationsTable extends React.Component {
         snackbar: PropTypes.object.isRequired,
         history: PropTypes.object.isRequired,
         loading: PropTypes.object.isRequired,
+    };
+
+    static model = {
+        modelValidations: {
+            href: { type: "URL" },
+            timestamp: { type: "DATE" },
+            metadata: { type: "COLLECTION" },
+        },
     };
 
     backHome = () => {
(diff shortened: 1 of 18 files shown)


src/components/notifications-table/NotificationsTable.jsx
```diff
@@ -0,0 +1,212 @@
...
+        this.setState({ toDelete: null });
+        this.props.loading.show(true, i18n.t("Deleting notification"));
+
+        await syncReport.remove(this.props.d2).then(response => {
+            this.props.loading.reset();
+
+            if (response.status) {
+                this.props.snackbar.success(i18n.t("Deleted {{name}}", { name: toDelete.name }));
+
+                // TODO: Workaround, add a way to force render of ObjectsTable on-demand
```
tokland: For the next meeting: we can add a specific prop, so instead of `key={tableKey}` we do `syncKey={syncKey}`. This way it's made explicit that Object Tables may and can be reloaded.

src/components/notifications-table/NotificationsTable.jsx
```diff
@@ -0,0 +1,212 @@
...
+        return (
+            <ul>
+                {namesToDisplay.map(name => (
+                    <li key={name}>{name}</li>
+                ))}
+            </ul>
+        );
+    };
+
+    // TODO: We should fix d2-ui-components instead
```
tokland: We'll see this on the meeting, bypassing the formatter was a way to get full control of how a field will be rendered.

src/components/notifications-table/NotificationsTable.jsx
```diff
@@ -0,0 +1,212 @@
...
+            <React.Fragment>
+                <ConfirmationDialog
+                    isOpen={!!toDelete}
+                    onSave={this.confirmDelete}
+                    onCancel={this.cancelDelete}
+                    title={i18n.t("Delete Instance?")}
+                    description={i18n.t("Are you sure you want to delete this instance?")}
+                    saveText={i18n.t("Ok")}
+                />
+                <PageHeader title={i18n.t("Notifications")} onBackClick={this.backHome} />
```
tokland: [minor and subjective] Do you like leaving a blank line between components at the same level?

src/types/synchronization.d.ts
```diff
@@ -21,15 +21,27 @@ export interface MetadataPackage {
...
+        typeStats: MetadataImportStats[];
         messages: any[];
     };
 }
 
+export type SynchronizationReportStatus = "READY" | "RUNNING" | "FAILURE" | "DONE";
+
+export interface SynchronizationReport {
+    id?: string;
+    timestamp?: Date;
```
tokland: Minor: Why don't we always set that field?

src/types/synchronization.d.ts
```diff
@@ -21,15 +21,27 @@ export interface MetadataPackage {
...
+    report?: {
+        typeStats: MetadataImportStats[];
         messages: any[];
     };
 }
 
+export type SynchronizationReportStatus = "READY" | "RUNNING" | "FAILURE" | "DONE";
+
+export interface SynchronizationReport {
+    id?: string;
```
tokland: I guess the id is left optional because there is no id when we use build. Not necessary here, just as a note, I looked at how to have types without some fields. build would use the id-less type. Example:

```
type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;

type Pos3D = { x: number; y: number; z: number };
type Pos2D = Omit<Pos3D, "z">;
type Pos1D = Omit<Pos3D, "y" | "z">;
```

src/utils/synchronization.ts
```diff
@@ -38,6 +38,11 @@ export function cleanReferences(
     return _.intersection(_.keys(references), rules);
 }
 
+export function createMetadataExportUrl(d2: D2, elements: string[]): string {
+    const requestUrl = d2.Api.getApi().baseUrl + "/metadata.json";
+    return requestUrl + `?fields=:all&defaults=EXCLUDE&filter=id:in:[${elements.join(",")}]`;
```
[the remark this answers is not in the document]
SferaDev: Yeah, for the real synchronization we slice the array in groups of 100 to be conservative (https://github.com/EyeSeeTea/metadata-synchronization/blob/development/src/utils/synchronization.ts#L45)

Not sure if we want to have multiple URLs here too?

src/models/syncReport.ts
```diff
@@ -0,0 +1,79 @@
...
+        const data = await getDataById(d2, notificationsDataStoreKey, id);
+        return this.build(data);
+    }
+
+    public static async list(
+        d2: D2,
+        filters: TableFilters,
+        pagination: TablePagination
+    ): Promise<TableList> {
+        return getPaginatedData(d2, notificationsDataStoreKey, filters, pagination);
```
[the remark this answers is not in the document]
SferaDev: See Reduce network calls of dataStore #83

src/types/synchronization.d.ts
SferaDev: Actually we're setting it now on the constructor, so we do not have to manually set it manually, same goes for the id.

SferaDev:
@tokland See the unresolved-conversations on the PR

I changed:

- tsconfig: "noUnusedParameters": true, "noUnusedLocals": true

- postMetadata: Now returns an AxiosResponse instead of any, note that I need to convert the object literal to any before returning (it's weird TS)

- hasErrors: Now returns the _.some(status !== OK) as we discussed

src/utils/synchronization.ts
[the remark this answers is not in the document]
tokland: I guess for the general case we should also split here in N links. It's cumbersome, but I don't see any other way.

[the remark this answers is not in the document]
SferaDev: I will add it on a new smaller PR later today

PR #80, opened 2019-05-21
Use same protocol for jquery load

tokland:
### :pushpin: References

Error: "Cannot read property feedbackDhis2 of undefined"

### :memo: Implementation

- `index.html` loaded jquery with `http://`, let's use `//` (same protocol)

https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content/How_to_fix_website_with_mixed_content

Note that this will be a problem if we need syncs to/from dhis2 instances on http from the app in a https instance.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -20,7 +20,7 @@
     -->
     <title>Metadata Synchronization</title>
 
-    <script src="http://code.jquery.com/jquery-latest.min.js"></script>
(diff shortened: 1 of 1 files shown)


adrianq:
Closing this in favour of https://github.com/EyeSeeTea/metadata-synchronization/pull/81

Issue #39, opened 2019-03-22
Improve Instances configuration


adrianq:
For later:
- [ ] Add cypress integration tests for creating a test instance
- [ ] Reduce number of calls creating namespace of dataStore: Remove check calls and add try/catch if 'GET' call fails

- [ ] Remove model [prop](https://github.com/EyeSeeTea/metadata-synchronization/blob/development/src/components/instance-configurator/InstanceConfigurator.jsx#L90) from instance-configurator
- [ ] Add dataStore end2end tests for adding new instances (to be discussed)
- [ ] Rename form-builder and use an abstract d2-ui-components version

adrianq:
Not relevant anymore. Instance configuration is not one of our major problems at the moment

PR #83, opened 2019-05-21
Reduce network calls of dataStore

SferaDev:
### :pushpin: References

* **Issue:** Work on #39 

### :memo: Implementation

- Asume namespace and dataStore exist and detect 404 errors

### DIFF
--- a/src/models/dataStore.ts
+++ b/src/models/dataStore.ts
@@ -16,7 +16,11 @@ async function getOrCreateNamespace(d2: D2): Promise<any> {
     }
 }
 
-async function getDataStore(d2: D2, dataStoreKey: string, defaultValue: any = []): Promise<any> {
+async function getDataStore(
+    d2: D2,
+    dataStoreKey: string,
+    defaultValue: any = undefined
+): Promise<any> {
     const dataStore = await getOrCreateNamespace(d2);
     try {
         return await dataStore.get(dataStoreKey);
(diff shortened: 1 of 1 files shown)


src/models/dataStore.ts
```diff
@@ -5,21 +5,25 @@ import { TableFilters, TableList, TablePagination } from "../types/d2-ui-compone
...
 async function getOrCreateNamespace(d2: D2): Promise<any> {
-    const existsNamespace = await d2.dataStore.has(dataStoreNamespace);
-    if (!existsNamespace) {
-        return await d2.dataStore.create(dataStoreNamespace);
-    } else {
+    try {
         return await d2.dataStore.get(dataStoreNamespace);
+    } catch (e) {
+        if (e.httpStatusCode === 404) return await d2.dataStore.create(dataStoreNamespace);
+        throw new Error(`Unhandled error with dataStore namespace ${dataStoreNamespace}`);
```
tokland: A full if/else is more verbose (4 lines instead of 2), but I think it's more clear.

src/models/dataStore.ts
```diff
@@ -5,21 +5,25 @@ import { TableFilters, TableList, TablePagination } from "../types/d2-ui-compone
...
     }
 }
 
 async function getDataStore(d2: D2, dataStoreKey: string, defaultValue: any = []): Promise<any> {
```
tokland: I guess that `[]` was a convenience so the callers are unchanged, but it's a bit weird to have a default value of array (typically: `undefined`).

src/models/dataStore.ts
```diff
@@ -5,21 +5,25 @@ import { TableFilters, TableList, TablePagination } from "../types/d2-ui-compone
...
-    if (!existsNamespace) {
-        await dataStore.set(dataStoreKey, defaultValue);
+    try {
+        return await dataStore.get(dataStoreKey);
+    } catch (e) {
+        if (e.httpStatusCode === 404) {
+            await dataStore.set(dataStoreKey, defaultValue);
+            return defaultValue;
+        }
+        throw new Error(`Unhandled error with dataStore ${dataStoreKey}`);
```
tokland: I am a bit wary of this kind of throws because we lose the original stack. What about simply rethrowing it, `throw e`?

PR #91, opened 2019-05-24
Improve performance on metadata synchronization

### LINKED ISSUE #86: Refine sync process performance
Now:
- [x] Remove metadata package (https://github.com/EyeSeeTea/metadata-synchronization/pull/75)
- [x] Remove type report (raw data) and keep report field (https://github.com/EyeSeeTea/metadata-synchronization/pull/75)

- [x] Store data already retrieved (attributes, groups and groups sets and any other potential dependency) and don't ask for them again (give a try to coroutines, use a block scoped dictionary, if concurrency needs to be addressed wrap the promises with p-limit) 
- [x] Update message in loading screen (generator approach with a machine of states)

SferaDev:
### :pushpin: References

* **Issue:** Closes #86 

### :memo: Implementation

- Transform synchronization into a state machine with generator
- Update loading page with messages
- Add a visited id dictionary and move recursive function to a inner method
- Add missing TS definitions for arrow methods

### 🔢 Performace

- Previous algorithm took linear time
- Current algorithm takes logarithmic time
- For detailed comparison see https://github.com/EyeSeeTea/metadata-synchronization/issues/86#issuecomment-495881754

### :art: Screenshots

image

image

image

image

image

### DIFF
--- a/src/components/sync-dialog/SyncDialog.jsx
+++ b/src/components/sync-dialog/SyncDialog.jsx
@@ -2,9 +2,8 @@ import React from "react";
 import i18n from "@dhis2/d2-i18n";
 import PropTypes from "prop-types";
 import _ from "lodash";
-import { ConfirmationDialog, MultiSelector } from "d2-ui-components";
+import { ConfirmationDialog, MultiSelector, withLoading } from "d2-ui-components";
 import DialogContent from "@material-ui/core/DialogContent";
-import { withLoading } from "d2-ui-components";
 
 import Instance from "../../models/instance";
 import { startSynchronization } from "../../logic/synchronization";
@@ -41,20 +40,25 @@ class SyncDialog extends React.Component {
     };
 
     handleSynchronization = async () => {
-        this.props.loading.show(true, i18n.t("Synchronizing metadata"));
+        const { handleClose, loading, metadata, d2 } = this.props;
+        loading.show(true, i18n.t("Synchronizing metadata"));
 
         try {
-            const syncReport = await startSynchronization(this.props.d2, {
-                metadata: this.props.metadata,
+            const builder = {
+                metadata: metadata,
                 targetInstances: this.state.targetInstances,
-            });
-            this.props.handleClose(syncReport);
+            };
+            for await (const { message, syncReport, done } of startSynchronization(d2, builder)) {
+                if (message) loading.show(true, message);
+                if (syncReport) await syncReport.save();
+                if (done) handleClose(syncReport);
+            }
         } catch (error) {
             console.error(error);
-            this.props.handleClose();
+            handleClose();
         }
         this.setState({ targetInstances: [] });
-        this.props.loading.reset();
+        loading.reset();
     };
 
     handleCancel = () => {
--- a/src/components/sync-dialog/SyncDialog.jsx
+++ b/src/components/sync-dialog/SyncDialog.jsx
@@ -41,12 +41,13 @@ class SyncDialog extends React.Component {
 
     handleSynchronization = async () => {
         const { handleClose, loading, metadata, d2 } = this.props;
+        const { targetInstances } = this.state;
         loading.show(true, i18n.t("Synchronizing metadata"));
 
         try {
             const builder = {
                 metadata: metadata,
-                targetInstances: this.state.targetInstances,
+                targetInstances: targetInstances,
             };
             for await (const { message, syncReport, done } of startSynchronization(d2, builder)) {
                 if (message) loading.show(true, message);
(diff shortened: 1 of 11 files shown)


src/components/sync-dialog/SyncDialog.jsx
```diff
@@ -41,20 +40,25 @@ class SyncDialog extends React.Component {
...
-        this.props.loading.show(true, i18n.t("Synchronizing metadata"));
+        const { handleClose, loading, metadata, d2 } = this.props;
+        loading.show(true, i18n.t("Synchronizing metadata"));
 
         try {
-            const syncReport = await startSynchronization(this.props.d2, {
-                metadata: this.props.metadata,
+            const builder = {
+                metadata: metadata,
                 targetInstances: this.state.targetInstances,
```
tokland: [Very minor] We could unpack `targetInstances` from state at the top of the method, that way we see in a glance what props/state this method deals with.

src/components/sync-dialog/SyncDialog.jsx
```diff
@@ -41,20 +40,25 @@ class SyncDialog extends React.Component {
...
-            });
-            this.props.handleClose(syncReport);
+            };
+            for await (const { message, syncReport, done } of startSynchronization(d2, builder)) {
+                if (message) loading.show(true, message);
+                if (syncReport) await syncReport.save();
+                if (done) handleClose(syncReport);
```
tokland: Minor: I guess `done` is implicit to the generator stopping, could we simply do the handleClose out of the for loop? Or is there a case where you yield done=true and later yield other values?

src/logic/synchronization.ts
```diff
@@ -105,57 +124,70 @@ async function importMetadata(
...
+        );
     const exportResults: MetadataPackage[] = await Promise.all(exportPromises);
     const metadataPackage = _.deepMerge({}, ...exportResults);
     console.debug("Metadata package from origin instance done", metadataPackage);
 
     // Phase 2: Create parent synchronization task
+    yield { message: i18n.t("Retrieving information from remote instances") };
     const targetInstances: Instance[] = await Promise.all(
-        targetInstanceIds.map(id => Instance.get(d2, id))
+        targetInstanceIds.map((id: string): Promise<Instance> => Instance.get(d2, id))
```
tokland: [minor] Do we need those explicit types? In a lambda, typically you onit the types if TS knows the types. There are some other instances of this in the code.

src/components/sync-dialog/SyncDialog.jsx
SferaDev: I added this inside so we do not have to store the last syncReport in a temporal local variable outside the for...of.

```
let lastSyncReport = undefined;
for await (const { message, syncReport, done } of startSynchronization(d2, builder)) {
                if (message) loading.show(true, message);
                if (syncReport) await syncReport.save();
                lastSyncReport = syncReport;
handleClose(lastSyncReport);
```

Also note that this done is not the implicit done of the Generator, during the for...of you loose the internal properties and only receive the value.

src/components/sync-dialog/SyncDialog.jsx
SferaDev: Sure!

src/logic/synchronization.ts
SferaDev: We had places with them, and others without. I just added them everywhere. No strong opinion about leaving or keeping them.

src/components/sync-dialog/SyncDialog.jsx
tokland: Yes, that's a custom `done` key in the object. Ok, I see your point.

src/logic/synchronization.ts
```diff
@@ -23,46 +25,63 @@ import {
...
+                        );
+                        return recursiveExport(newBuilder);
+                    }
+                );
+            const promisesResult: MetadataPackage[] = await Promise.all(promises);
+            _.deepMerge(result, ...promisesResult);
+        }
+
+        // Clean up result from duplicated elements
+        _.forOwn(
```
tokland: `result` needs to be updated imperatively? Functional approach:

return _.mapValues(result, (objects, metadataType) => _.uniqBy(objects, "id"))

adrianq:
@SferaDev if you can also merge development into this branch it would be great. I have already merged the latest hotfixes in master to development and bump the version. thanks!

adrianq:
@SferaDev It works like a charm. I have found some minor issues but I am merging this PR and we can address them in another PR. See as follow
- When deleting a notification modal dialog displays wrong title and body (https://github.com/EyeSeeTea/metadata-synchronization/issues/101)
- What do we do with bulk actions? Do we add 'delete' to bulk actions? (https://github.com/EyeSeeTea/metadata-synchronization/issues/101)
- I have got a few Failures when syncing an OU (you can have a look to the notifications at dev2:8083) It was a bit strange because there was no reason for those syncs to fail and they look more like a warning than a failure. I haven't been able to reproduce the error or define when it was actually happening. Maybe you can have a look at the notifications and try to guess what is going on...

PR #95, opened 2019-05-29
Add Notification dropdown by status

### LINKED ISSUE #96: Add filter by status (dropdown) in Notifications page

SferaDev:
### :pushpin: References

* **Issue:** Closes #96 

### :memo: Implementation

- Notifications Page: Add filter by status in a dropdown


i18n/en.pot
```diff
@@ -146,6 +146,21 @@ msgstr ""
...
+msgstr ""
+
+msgid "Failure"
+msgstr ""
+
+msgid "Done"
+msgstr ""
+
+msgid "Synchronization Status"
+msgstr ""
```
tokland: [Minor] We show lowercase here but uppercase in the table:

image

tokland: Also, the dropdown is very close to the title. This didn't happen before because we had the search bar. Not sure if this should/can be fixed.

src/components/notifications-table/NotificationsTable.jsx
```diff
@@ -141,8 +145,50 @@ class NotificationsTable extends React.Component {
...
+            {
+                id: "FAILURE",
+                name: i18n.t("Failure"),
+            },
+            {
+                id: "DONE",
+                name: i18n.t("Done"),
+            },
+        ];
+        this.setState({ statusFilter: { value, items } });
```
tokland: All info is know, can we move it to the initial code?

i18n/en.pot
SferaDev: We noticed with Adrian too the spacing, will add some margin to top.

src/components/notifications-table/NotificationsTable.jsx
SferaDev: Yes! I mirrored the logic of the filters in BaseSyncConfigurator where the items are loaded async 😑

SferaDev:
Fixed comments from pull request:

- Added the missing d2, it's true it's solved on another PR but doesn't hurt to merge it now.

- The ObjectsTable now shows errors properly "Done" instead of "DONE".

- Added some spacing to top.

- Moved the dropdown value directly to state and items to internal prop

SferaDev:
@adrianq Yesterday you complained that we were getting "FAILURE" when the sync had "WARNINGS", in this PR we also fix that by only setting the "FAILURE" flag with straight errors (DHIS2 errors or network errors).

PR #98, opened 2019-05-29
Refactor dataStore for syncReports

### LINKED ISSUE #97: Refactor dataStore for syncReports
- [x] Refactor data store:
   - [x] Keep in http://dev2.eyeseetea.com:8083/api/dataStore/metatada-synchronization/notifications only id, user, timestamp, status and selectedTypes. Short labels, example user -> u
   - [x] Move results to each own key, example:http://dev2.eyeseetea.com:8083/api/dataStore/metatada-synchronization/notification-[id]

Future: 
- [ ] Do not wait for notification sent to dataStore, implement a mutex to prevent data races. 
- [ ] Keep track of all the status in a sync process. Two approaches: different records in data store that are grouped in the GUI (note this won't work easily if we implement buckets) or a single record in data store.

SferaDev:
### :pushpin: References

* **Issue:** Work #97 

### :memo: Implementation

- Fix typo introduced in #21 

- Instead of relying on d2 library to make the dataStore calls, use it directly through axios (this reduces the network calls for namespace existence).

- Rename selectedTypes to types and timestamp to date.

- Store results to their own dataStore namespace ```notification-${id}```

### :bookmark_tabs: Others

There are multiple approaches to implement this functionality.

These are the design choices made:

- We have separated the syncReport and results objects and modified the save/remove methods to update the new namespace.

- To avoid loading more information than needed, the results object will be set to NULL in most cases, the SyncSummary loads it on-demand during componentDidUpdate.


src/components/sync-summary/SyncSummary.jsx
```diff
@@ -185,14 +186,7 @@ class SyncSummary extends React.Component {
...
                             <ExpansionPanelDetails>
-                                <ReactJson
-                                    src={{
-                                        results: response.results,
-                                        syncReport: response.syncReport,
-                                    }}
-                                    collapsed={2}
-                                    enableClipboard={false}
-                                />
+                                <ReactJson src={response} collapsed={2} enableClipboard={false} />
```
SferaDev: The problem here is that we print injected internal class properties that should be hidden (as an example selected property from ObjectsTable). I thought of adding a lodash pick of results and syncReport. What do you think?

tokland: I see! Then, yes, we need to recover a `_.pick` here.

SferaDev: Will update the PR in a few minutes so it's ready to merge

adrianq:
@SferaDev It works nicely! Just a minor comment, we can maybe start with the Object Table sorting by Timestamp showing the most recent one at the beginning. What do u think?

SferaDev:
> @SferaDev It works nicely! Just a minor comment, we can maybe start with the Object Table sorting by Timestamp showing the most recent one at the beginning. What do u think?

Sure thing, see #104

PR #108, opened 2019-06-12
Refactor pages and components

SferaDev:
### :pushpin: References

* **Issue:** Work #39 

### :memo: Implementation

- Rename components with more explicit naming (ie: pages)

- Clean-up imports and re-order methods/properties

- Fix couple of async return await

- Move themes to ./components/app

- Since we've just tagged a release, now it's the perfect moment for this refactor before adding the sync rules (which adds 2 more pages with an ObjectsTable, Wizard, ...)

### :bookmark_tabs: Others

- To avoid complexity during the review of the sync rules feature, this PR only includes renaming/moving of files.


i18n/en.pot
```diff
@@ -5,143 +5,135 @@ msgstr ""
...
 
-msgid "Instance Configurator"
-msgstr ""
-
-msgid "Organisation Units Sync"
+msgid "<No value>"
 msgstr ""
 
-msgid "Data Elements Sync"
+msgid "Server name (*)"
```
tokland: [Note: not for this PR] Maybe have translations without the `(*)` required field suffix and add them programmatically whenever needed.

src/models/instance.ts
```diff
@@ -31,6 +31,30 @@ export default class Instance {
...
+    public get username(): string {
+        return this.data.username;
+    }
+
+    public get password(): string {
+        return this.data.password;
+    }
+
+    public get description(): string {
+        return this.data.description ? this.data.description : "";
```
tokland: [ultra-minor] [Note: not for this refactor PR] `x ? x : y` -> `x || y`.

SferaDev: Yeah, I tend to forget this exists! Thanks @tokland

PR #111, opened 2019-06-27
Improve landing page look and feel

SferaDev:
### :memo: Implementation

- Follow vaccination's landing page with a Grid that hosts several rows/columns

### :art: Screenshots

image


src/components/landing-page/LandingPage.jsx
```diff
@@ -3,17 +3,49 @@ import PropTypes from "prop-types";
...
+    },
+    item: {
+        padding: 20,
+    },
+    title: {
+        marginLeft: 10,
+    },
+    icons: {
+        fontSize: "70px !important",
+        marginRight: 10,
```
tokland: Depending on the zoom-level, elements in the grid may have different heights. Is this a problem? (or can it be solved?)

image

PR #117, opened 2019-07-01
Allow synchronization of OrganisationUnitGroups

### LINKED ISSUE #115: Implement Organisation Unit Group Sync

SferaDev:
### :pushpin: References

* **Issue:** Closes #115 

### :memo: Implementation

- Re-use our powerful core logic for dependency fetching
- Re-use the new abstracted Metadata Table from Sync Rules

### :art: Screenshots

image

- We avoid including ou/groups not related to the selected group

### :bookmark_tabs: Others

Example of sync for "Bethal Management Area"

```json
{
    "organisationUnitGroups": [
        {
            "code": "OUG_BETHAL_MA",
            "id": "Gi0gY8zEjUe",
            "name": "Bethal Management Area",
            "organisationUnits": [
                {
                    "id": "aC2YG4xbDKt"
                },
                {
                    "id": "WbG7fgnRESE"
                },
                {
                    "id": "hflkMOWcKvp"
                },
                {
                    "id": "l27KmxnFQvF"
                }
            ],
            "groupSets": [
                {
                    "id": "OFVuC1nVGOA"
                }
            ]
        }
    ],
    "organisationUnits": [
        {
            "id": "aC2YG4xbDKt",
            "level": 5,
            "name": "mp Bethal Correctional Centre",
        },
        {
            "id": "hflkMOWcKvp",
            "level": 5,
            "name": "mp Ermelo Correctional Centre",
        },
        {
            "id": "WbG7fgnRESE",
            "level": 5,
            "name": "mp Piet Retief Correctional Centre",
        },
        {
            "id": "l27KmxnFQvF",
            "level": 5,
            "name": "mp Volksrust Correctional Centre",
        }
    ],
    "organisationUnitGroupSets": [
        {
            "code": "OUGS_DCS_MX_AREA",
            "id": "OFVuC1nVGOA",
            "name": "DCS Management Area",
        }
    ]
}
```


src/components/synchronization-page/OrganisationUnitPage.jsx
```diff
@@ -15,7 +15,11 @@ export default class OrganisationUnitPage extends React.Component {
         const title = i18n.t("Organisation Units Synchronization");
 
         return (
-            <GenericSynchronizationPage d2={d2} models={[OrganisationUnitModel]} title={title} />
+            <GenericSynchronizationPage
+                d2={d2}
+                models={[OrganisationUnitModel, OrganisationUnitGroupModel]}
```
tokland: Following the rule of thumb of not using created-on-the-fly values for props, we can use use a `this.models` like we do in `OrganisationUnitPage`.

adrianq:
It is really nice to have this abstract table! Thanks for this!

I think the workflow is a bit confusing for the user though. It looks like sync is only happening for the object selected and **displayed** on the screen. IMO this is a bit confusing. For instance, if I select 1 OU and 2 OUGs but I only displaying OUGs (because I am filtering by OUGs in the Metadata Type dropdown) synchronization will only happen for the OUGs and the other way around, if I am displaying only OU synchronization will only happen for OU. This is even more confusing if we compare it with sync rules. In the sync rule section, everything that is selected is part of the sync rule even if it is hidden because of a filter or whatever.

Finally, if I want to sync object with different metadata types I am getting this error:
image

I am merging this anyway because now that I understand the limitations I can demo it to the client (avoiding the conflicted use cases) but we need to rethink this a bit

SferaDev:
> It is really nice to have this abstract table! Thanks for this!
> 
> I think the workflow is a bit confusing for the user though. It looks like sync is only happening for the object selected and **displayed** on the screen. IMO this is a bit confusing. For instance, if I select 1 OU and 2 OUGs but I only displaying OUGs (because I am filtering by OUGs in the Metadata Type dropdown) synchronization will only happen for the OUGs and the other way around, if I am displaying only OU synchronization will only happen for OU. This is even more confusing if we compare it with sync rules. In the sync rule section, everything that is selected is part of the sync rule even if it is hidden because of a filter or whatever.
> 
> Finally, if I want to sync object with different metadata types I am getting this error:
> image
> 
> I am merging this anyway because now that I understand the limitations I can demo it to the client (avoiding the conflicted use cases) but we need to rethink this a bit

This is in fact an error, the sync button is setting somewhere the type then. I will review it so we can fix it before the release on Friday.

PR #120, opened 2019-07-02
Sync rules improvements

### LINKED ISSUE #119: Sync Rule Improvements
- [x] List of target instances should display as separated commas list
- [x] Left click on a row forward to the edit feature. While in the other sync section it goes to Details. Although in latest versions of DHIS2 the first option is the implemented one, in general, unless you have a strong opinion in a different sense I will choose 'Details' as the left click action.
- [x] Remove asterisk and use compulsory fields instead ([reference](https://github.com/EyeSeeTea/metadata-synchronization/pull/108#discussion_r294640352))
- [ ] If you click on 'Clear Selection' probably is better to clear up all the filters (including the selected checkbox) and refresh, right?

SferaDev:
### :pushpin: References

* **Issue:** Closes #119 

### :memo: Implementation

- Add asterisks programatically in i18n strings
- Make target instances a csv in Sync Rules
- Make default action to be edit

### :art: Screenshots

image

image

image

### DIFF
--- a/src/components/instance-creation-page/GeneralInfoForm.jsx
+++ b/src/components/instance-creation-page/GeneralInfoForm.jsx
@@ -81,7 +81,7 @@ class GeneralInfoForm extends React.Component {
                 value: instance.name,
                 component: TextField,
                 props: {
-                    floatingLabelText: i18n.t("Server name (*)"),
+                    floatingLabelText: `${i18n.t("Server name")} (*)`,
                     style: { width: "100%" },
                     changeEvent: "onBlur",
                     "data-test": "name",
@@ -112,7 +112,7 @@ class GeneralInfoForm extends React.Component {
                 value: instance.url,
                 component: TextField,
                 props: {
-                    floatingLabelText: i18n.t("URL endpoint (*)"),
+                    floatingLabelText: `${i18n.t("URL endpoint")} (*)`,
                     style: { width: "100%" },
                     changeEvent: "onBlur",
                     "data-test": "url",
@@ -137,7 +137,7 @@ class GeneralInfoForm extends React.Component {
                 value: instance.username,
                 component: TextField,
                 props: {
-                    floatingLabelText: i18n.t("Username (*)"),
+                    floatingLabelText: `${i18n.t("Username")} (*)`,
                     style: { width: "100%" },
                     changeEvent: "onBlur",
                     "data-test": "username",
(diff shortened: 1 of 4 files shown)


SferaDev:
@adrianq said:
>  If you click on 'Clear Selection' probably is better to clear up all the filters (including the selected checkbox) and refresh, right?

This is not particularly easy to follow without adding more complexity to d2-ui-components ObjectsTable.

If we implement it in the application space only with the selection has changed method, we will reset the filters with every selectedIds is empty. (Meaning selecting 1, deseleting 1, will incur in forgetting all the filters).

Updating issue https://github.com/EyeSeeTea/d2-ui-components/issues/10 to be done in the future

src/components/instance-creation-page/GeneralInfoForm.jsx
```diff
@@ -81,7 +81,7 @@ class GeneralInfoForm extends React.Component {
                 value: instance.name,
                 component: TextField,
                 props: {
-                    floatingLabelText: i18n.t("Server name (*)"),
```
tokland: Is it worth to create some `function optional(s) { return `${s} (*)`; } in some utils file to abstract it? Not sure.

src/components/rules-list-page/SyncRulesPage.jsx
```diff
@@ -42,35 +43,44 @@ class SyncRulesPage extends React.Component {
...
+            getValue: ruleData => {
+                const { allInstances } = this.state;
+                const rule = SyncRule.build(ruleData);
+                return getValueForCollection(
+                    _(rule.targetInstances)
+                        .map(id => allInstances.find(instance => instance.id === id))
+                        .compact()
+                        .map(({ name }) => ({ name }))
+                );
+            },
```
tokland: Those two `getValue` share some code. Maybe we can abstract it to a method. Something like: 

```
            getTargetInstances() {
                const { allInstances } = this.state;
                const rule = SyncRule.build(ruleData);
                return _(rule.targetInstances)
                    .map(id => allInstances.find(instance => instance.id === id))
                    .compact()
                    .map(({ name }) => ({ name }))
                );
            }
```

SferaDev: Yeah sounds good

src/components/instance-creation-page/GeneralInfoForm.jsx
SferaDev: Both options are fine

adrianq:
No worries about the 'Clear Selection' issue is a very minor issue. PR seems to be working fine

PR #121, opened 2019-07-02
Remove user related properties for all types

SferaDev:
### :pushpin: References

- @tokland Found an issue trying to sync a DE and was because we forgot to add an exlude rule for ```dataElement.dataSet.userGroupAccesses```.

### :memo: Implementation

- Instead of having to add exclude rules for-each metadata type, create a global rule that removes those properties from being synchronized.

- We do not have any scenario yet where we want to migrate those properties


SferaDev:
> I am so familiar with the dependency tree to know all the implications of removing those keys (I'll leave this to you ;)), but the POST now works and code looks ok, so fine by me!

It completely depends on the master-slave instance set-up, we didn't originally create a global rule because we were waiting for feedback on past milestone releases. But since we do not have specific feedback yet and to avoid crashes, it's good to leave it this way.

adrianq:
Works fine and I definitely think is a good idea to have some global exclude rules!

PR #127, opened 2019-07-09
[Refactor] Clean-up several minor issues

SferaDev:
### :memo: Implementation

- Remove TODO console.log message on landing page

- Rename Synchronization History page

- Rename Instance Configuration page

- Make code consistent with the current naming scheme

### :bookmark_tabs: Others

- Not sure about renaming the dataStore namespace from ```notifications``` to ```history``` as it's a breaking change (notifications already created will vanish).


adrianq:
All good here. I am fine with no changing the name in the data store. Probably it will cause more problems than improvements so to say

PR #128, opened 2019-07-09
[History] Visualize if a sync was executed by a rule

### LINKED ISSUE #113: Modify notification page to store sync rule id if any
- [x] Add sync rule id to the notification
- [x] Add new column with sync rule name
- [x] Add link to sync rule edit in details panel

SferaDev:
### :pushpin: References

* **Issue:** Closes #113 

### :memo: Implementation

- Add support for ```paging=false``` in dataStore getPaginatedData()
- Add syncRule id to a syncReport if the sync was executed as part of a rule
- Show the syncRule in the Notifications page

### :art: Screenshots

image

image

### :bookmark_tabs: Others

- https://github.com/EyeSeeTea/d2-ui-components/pull/50

### DIFF
--- a/src/components/history-list-page/HistoryPage.jsx
+++ b/src/components/history-list-page/HistoryPage.jsx
@@ -3,12 +3,13 @@ import PropTypes from "prop-types";
 import _ from "lodash";
 import i18n from "@dhis2/d2-i18n";
 import { ConfirmationDialog, ObjectsTable, withLoading, withSnackbar } from "d2-ui-components";
-import { withRouter } from "react-router-dom";
+import { Link, withRouter } from "react-router-dom";
 import { withStyles } from "@material-ui/core/styles";
 
 import PageHeader from "../page-header/PageHeader";
 import Dropdown from "../dropdown/Dropdown";
 import SyncReport from "../../models/syncReport";
+import SyncRule from "../../models/syncRule";
 import SyncSummary from "../sync-summary/SyncSummary";
 import { getValueForCollection } from "../../utils/d2-ui-components";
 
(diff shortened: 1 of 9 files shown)


src/components/history-list-page/HistoryPage.jsx
```diff
@@ -156,6 +158,34 @@ class HistoryPage extends React.Component {
             text: i18n.t("Metadata Types"),
             getValue: notification => this.getMetadataTypes(notification),
         },
+        {
+            name: "syncRule",
+            text: i18n.t("Sync Rule"),
+            getValue: ({ syncRule }) => this.getSyncRuleEditLink(syncRule),
```
tokland: When checking the details of a non-rule sync, we see the placeholder "-" value for the "Sync Rule" attribute, which is a bit ugly. I see that I provided no way of hidden the field:

https://github.com/EyeSeeTea/d2-ui-components/blob/master/src/details-box/DetailsBox.component.js#L30

Not necessarily for now, but on the future it would be cool to simply don't render the field if `getValue` returns `undefined`. For now, we can return here `undefined` here and some time make the change in `DetailsBox`.

src/components/rules-creation-page/steps/InstanceSelectionStep.jsx
```diff
@@ -4,12 +4,8 @@ import { MultiSelector } from "d2-ui-components";
 import Instance from "../../../models/instance";
 
 export const getInstances = async d2 => {
-    const instances = await Instance.list(
-        d2,
-        { search: "" },
-        { page: 1, pageSize: 100, sorting: [] }
-    );
-    return instances.objects.map(instance => ({
+    const { objects } = await Instance.list(d2, null, { paging: false });
```
tokland: [minor] The signature is `list(d2: D2, filters: TableFilters, pagination: TablePagination)`, we can pass `{}` as `filters`.

src/components/history-list-page/HistoryPage.jsx
```diff
@@ -143,6 +134,17 @@ class HistoryPage extends React.Component {
         return getValueForCollection(notification.types.map(type => ({ name: type })));
     };
 
+    getSyncRuleEditLink = id => {
+        const { syncRules } = this.state;
+        const syncRule = syncRules.find(e => e.id === id);
```
tokland: ultra-minor: what does `e` stand for? :)

src/components/history-list-page/HistoryPage.jsx
SferaDev: What about:

- getValue: render getValue()
- primitive: print primite value
- null: Hide property
- undefined: print "-"

This way you can still print "-" for descriptions, and when you actually want to hide it you set it to null.

tokland: LGTM!

PR #130, opened 2019-07-13
[Scheduler] Add new scheduler to Sync Rules

### LINKED ISSUE #112: Modify sync rules to include scheduler
- [x] Add two columns ('Frequency' and 'Scheduled') in sync rule landing page
- [x] Add new step (4) 'Schedule'
    - Reuse DHIS2 Scheduler interface
    - This step must be optional
    - Store the scheduled date to DHIS2
    - Fields: Select frequency, cron expressions and enable/disabled
- [x] Add new contextual action to enable/disable. If no scheduler we will display an error message.

References:

- https://github.com/dhis2/scheduler-app/blob/master/src/utils/validCronExpression.js

SferaDev:
### :pushpin: References

* **Issue:** Closes #112 

* This PR adds the basic functionality to stop blocking Phase 2 development

* UI/UX improvements should be implemented in upcoming PRs (example #131)

### :memo: Implementation

- Add new scheduling step to Sync Rules Wizard

- Fix general info step to be fully immutable

- Add validation before saving on Wizard

- Add new columns to Sync Rules Page to reflect scheduling

- Add cron expressions validation from @dhis2/scheduler-app

### :art: Screenshots

image

image

image

image

image

- Since we are using the dataStore as database and we are unable to store booleans, we do the string conversion on the model.

### DIFF
--- a/src/components/rules-creation-page/SyncRulesWizard.jsx
+++ b/src/components/rules-creation-page/SyncRulesWizard.jsx
@@ -9,6 +9,7 @@ import GeneralInfoStep from "./steps/GeneralInfoStep";
 import SyncRule from "../../models/syncRule";
 import MetadataStep from "./steps/MetadataSelectionStep";
 import InstanceSelectionStep from "./steps/InstanceSelectionStep";
+import SchedulerStep from "./steps/SchedulerStep";
 import SaveStep from "./steps/SaveStep";
 import { getValidationMessages } from "../../utils/validations";
 
@@ -41,6 +42,14 @@ class SyncRulesWizard extends React.Component {
             component: InstanceSelectionStep,
             validationKeys: ["targetInstances"],
             description: undefined,
+            help: undefined,
+        },
+        {
+            key: "scheduler",
+            label: i18n.t("Scheduling"),
+            component: SchedulerStep,
+            validationKeys: ["frequency", "enabled"],
+            description: i18n.t("Configure the scheduling frequency for the synchronization rule"),
             help: undefined,
         },
         {
--- a/src/components/rules-creation-page/steps/SaveStep.jsx
+++ b/src/components/rules-creation-page/steps/SaveStep.jsx
@@ -2,11 +2,12 @@ import React, { useEffect, useState } from "react";
 import _ from "lodash";
 import PropTypes from "prop-types";
 import { Button, LinearProgress, withStyles } from "@material-ui/core";
-import { ConfirmationDialog } from "d2-ui-components";
+import { ConfirmationDialog, withSnackbar } from "d2-ui-components";
 import i18n from "@dhis2/d2-i18n";
 
 import { getInstances } from "./InstanceSelectionStep";
 import { getMetadata } from "../../../utils/synchronization";
+import { getValidationMessages } from "../../../utils/validations";
 
 const LiEntry = ({ label, value, children }) => {
     return (
@@ -27,8 +28,7 @@ const styles = () => ({
     },
 });
 
-const SaveStep = props => {
-    const { d2, syncRule, classes, onCancel } = props;
+const SaveStep = ({ d2, syncRule, classes, onCancel, snackbar }) => {
     const [cancelDialogOpen, setCancelDialogOpen] = useState(false);
     const [isSaving, setIsSaving] = useState(false);
     const [metadata, updateMetadata] = useState({});
@@ -40,9 +40,18 @@ const SaveStep = props => {
 
     const save = async () => {
         setIsSaving(true);
-        await syncRule.save(d2);
+
+        const errors = await getValidationMessages(d2, syncRule);
+        if (errors.length > 0) {
+            snackbar.error(errors.join("\n"), {
+                autoHideDuration: null,
+            });
+        } else {
+            await syncRule.save(d2);
+            onCancel();
+        }
+
         setIsSaving(false);
-        onCancel();
     };
 
     useEffect(() => {
@@ -95,6 +104,10 @@ const SaveStep = props => {
                         })}
                     </ul>
                 </LiEntry>
+
+                <LiEntry label={i18n.t("Scheduling enabled")} value={syncRule.enabled.toString()} />
+
+                <LiEntry label={i18n.t("Frequency")} value={syncRule.frequency} />
             </ul>
 
             <Button onClick={openCancelDialog} variant="contained">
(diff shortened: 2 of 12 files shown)


src/utils/apiTranslations.js
```diff
@@ -0,0 +1,13 @@
...
+    en: {
+        // Cron expressions
+        CUSTOM: "Custom",
+        EVERY_DAY_MIDNIGHT: "Every day at midnight",
+        EVERY_DAY_THREE_AM: "Every day at 3 AM",
+        EVERY_HOUR: "Every hour",
+        EVERY_WEEK: "Every week",
+        EVERY_WEEKDAY_NOON: "Every day at noon",
+    },
+};
```
tokland: I guess that's because those translations where not appearing in i18n/en.pot. That's because the step has this: `displayName: i18n.t(text)`. Unfortunately, we cannot do that, the i18n extractor script expects the literal there. 

Proposal: Write i18n.t(...) in the component. Caveat: AFAIK, you cannot do it in the module definition (or class definition), when the module is loaded, the locale is not set yet. So you have to put it inside a function or similar. It's not trivial to test in this app because we only have EN.

src/components/rules-creation-page/SyncRulesWizard.jsx
```diff
@@ -43,6 +44,14 @@ class SyncRulesWizard extends React.Component {
             description: undefined,
             help: undefined,
         },
+        {
+            key: "scheduler",
+            label: i18n.t("Scheduling"),
+            component: SchedulerStep,
+            validationKeys: ["frequency", "enabled"],
+            description: i18n.t("Configure the scheduling frequency for the synchronization rule"),
```
tokland: Maybe add a link to documentation about the format of the cron expression? In vaccination-app we use react-linkify to render links easily, but feel free to do it how you see fit (if we really need a link here!)

src/utils/validCronExpression.js
```diff
@@ -0,0 +1,85 @@
...
+    if (!exp) {
+        return false;
+    }
+
+    const fields = exp.trim().split(" ");
+    if (!isValidFields(fields)) {
+        return false;
+    }
+
+    return (
```
tokland: [Subjective]

```
const fields = (exp || "").trim().split(" ");
return (
  isValidFields(fields) &&
 ...same..
)
```

src/utils/validCronExpression.js
```diff
@@ -0,0 +1,85 @@
...
+        isValidSecondField(fields[0]) &&
```
tokland: [Subjective] I am a fan of destructuring arrays to avoid non-meaningful number indexation. const { second, hour, .... } = fields;

src/models/syncRule.ts
```diff
@@ -102,6 +125,20 @@ export default class SyncRule {
         });
     }
 
+    public updateEnabled(enabled: boolean): SyncRule {
+        return SyncRule.build({
+            ...this.syncRule,
+            enabled: enabled.toString(),
```
tokland: You said in the description that booleans cannot be used, and indeed there is no type boolean in the UI of Dhis2. But is this limitation real? at least in user data store I am using booleans with no problems.

src/components/rules-creation-page/steps/SchedulerStep.jsx
```diff
@@ -0,0 +1,100 @@
...
+    { text: "EVERY_DAY_THREE_AM", value: "0 0 3 ? * *" },
+    { text: "EVERY_WEEKDAY_NOON", value: "0 0 12 ? * MON-FRI" },
+    { text: "EVERY_WEEK", value: "0 0 3 ? * MON" },
+];
+
+const Toggle = ({ label, onChange, value }) => (
+    <FormControlLabel
+        control={
+            <Switch
+                onChange={e => onChange(_.set(_.clone(e), "target.value", e.target.checked))}
```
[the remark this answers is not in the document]
SferaDev: Okay, looks good to me

src/components/rules-creation-page/steps/SaveStep.jsx
```diff
@@ -40,9 +40,18 @@ const SaveStep = props => {
 
     const save = async () => {
         setIsSaving(true);
-        await syncRule.save(d2);
+
+        const errors = await getValidationMessages(d2, syncRule);
+        if (errors.length > 0) {
+            snackbar.error(errors.join("\n"), {
+                autoHideDuration: null,
```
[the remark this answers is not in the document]
SferaDev: Yeah, will remove the prop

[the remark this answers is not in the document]
SferaDev: Yeah, will remove the prop

src/components/rules-creation-page/SyncRulesWizard.jsx
SferaDev: I think it's a good idea to add a link, but not sure to what page we should reference about cron expressions. We only support a subset of them (6 values including seconds) to make it more simple.

src/utils/validCronExpression.js
SferaDev: This is copy pasted code from ```scheduler-app```. I noticed some improvements too but left them as-is.

src/utils/validCronExpression.js
SferaDev: This is copy pasted code from ```scheduler-app```. I noticed some improvements too but left them as-is.

tokland: Oh, I see, then let's keep it as it is.

src/components/rules-creation-page/steps/SaveStep.jsx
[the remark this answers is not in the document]
SferaDev: Fixed on https://github.com/EyeSeeTea/metadata-synchronization/pull/134/commits/378f9bd18b6c4ea1ffd8763e047c6328049569b0

src/components/rules-creation-page/steps/SchedulerStep.jsx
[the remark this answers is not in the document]
SferaDev: Fixed on https://github.com/EyeSeeTea/metadata-synchronization/pull/134/commits/378f9bd18b6c4ea1ffd8763e047c6328049569b0

src/utils/apiTranslations.js
SferaDev: Fixed on https://github.com/EyeSeeTea/metadata-synchronization/pull/134/commits/378f9bd18b6c4ea1ffd8763e047c6328049569b0

SferaDev:
To avoid conflicts now with the merging. Some of the comments are fixed in #134 (the last PR we have now).

Also to not block the merging of these PRs, I've created #135 to keep track of pending changes. Will take a look on Monday to those.

adrianq:
It works like a charm. Very minor comment:
Maybe we can be a bit more explicit when there is no Frequency. Something like a '-' or 'No Frequency' should do the trick
image

SferaDev:
> It works like a charm. Very minor comment:
> Maybe we can be a bit more explicit when there is no Frequency. Something like a '-' or 'No Frequency' should do the trick

@adrianq This behavior changed a little bit on the last PR #135. Now if the cron expression is empty will not appear.

image

image

PR #132, opened 2019-07-15
[Scheduler] User friendly cron expressions

SferaDev:
### :memo: Implementation

- In the sync rules wizard, when inserting a valid cron expression, we previsualize the result. If it's incorrect we have the default "Select frequency" placeholder.

- We should probably update the options we suggest in the dropdown to match the same auto generated string (but I guess we should think before which cron expressions are useful to the clients ie: Every day at 11PM from Monday to Friday)

### :art: Screenshots

image

image

image

image

image

- The library supports multiple locales, not sure if we want to enable that feature yet (they support more languages than us and could lead to mixed translations UI)

- Added as string interpolation, also tested adding tooltips but not sure what's the best approach.

image

### DIFF
--- a/src/components/rules-creation-page/steps/SaveStep.jsx
+++ b/src/components/rules-creation-page/steps/SaveStep.jsx
@@ -9,6 +9,7 @@ import { ConfirmationDialog, withSnackbar } from "d2-ui-components";
 import { getInstances } from "./InstanceSelectionStep";
 import { getMetadata } from "../../../utils/synchronization";
 import { getValidationMessages } from "../../../utils/validations";
+import isValidCronExpression from "../../../utils/validCronExpression";
 
 const LiEntry = ({ label, value, children }) => {
     return (
@@ -108,10 +109,12 @@ const SaveStep = ({ d2, syncRule, classes, onCancel, snackbar }) => {
 
                 <LiEntry label={i18n.t("Scheduling enabled")} value={syncRule.enabled.toString()} />
 
-                <LiEntry
-                    label={i18n.t("Frequency")}
-                    value={`${cronstrue.toString(syncRule.frequency)} (${syncRule.frequency})`}
-                />
+                {isValidCronExpression(syncRule.frequency) && (
+                    <LiEntry
+                        label={i18n.t("Frequency")}
+                        value={`${cronstrue.toString(syncRule.frequency)} (${syncRule.frequency})`}
+                    />
+                )}
             </ul>
 
             <Button onClick={openCancelDialog} variant="contained">
--- a/src/components/rules-creation-page/steps/SaveStep.jsx
+++ b/src/components/rules-creation-page/steps/SaveStep.jsx
@@ -1,10 +1,10 @@
 import React, { useEffect, useState } from "react";
 import _ from "lodash";
 import PropTypes from "prop-types";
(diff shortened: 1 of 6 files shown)


src/components/rules-creation-page/steps/SchedulerStep.jsx
```diff
@@ -56,7 +57,11 @@ const SchedulerStep = ({ syncRule, onChange }) => {
             value: selectedCron.value || "",
             component: DropDown,
             props: {
-                hintText: i18n.t("Select frequency"),
+                hintText: i18n.t(
+                    isValidCronExpression(syncRule.frequency)
+                        ? cronstrue.toString(syncRule.frequency)
+                        : "Select frequency"
```
tokland: I'd keep the `i18n.t` here.

src/components/rules-creation-page/steps/SaveStep.jsx
```diff
@@ -107,7 +109,12 @@ const SaveStep = ({ d2, syncRule, classes, onCancel, snackbar }) => {
 
                 <LiEntry label={i18n.t("Scheduling enabled")} value={syncRule.enabled.toString()} />
 
-                <LiEntry label={i18n.t("Frequency")} value={syncRule.frequency} />
+                {isValidCronExpression(syncRule.frequency) && (
+                    <LiEntry
+                        label={i18n.t("Frequency")}
+                        value={`${cronstrue.toString(syncRule.frequency)} (${syncRule.frequency})`}
```
tokland: This snippet seens quite generic and it's repeated in the code. Maybe we can move it as a method of `SyncRule`?

src/components/rules-list-page/SyncRulesPage.jsx
```diff
@@ -67,20 +69,34 @@ class SyncRulesPage extends React.Component {
                     .map(e => e.name)
                     .join(", "),
         },
-        { name: "frequency", text: i18n.t("Frequency"), sortable: true },
+        {
+            name: "frequency",
+            text: i18n.t("Frequency"),
+            sortable: true,
+            getValue: ({ frequency }) =>
+                isValidCronExpression(frequency)
```
tokland: Same than before, does it make sense to have `syncRule.isValidFrequency()` or something like that?

src/components/rules-creation-page/steps/SchedulerStep.jsx
SferaDev: See https://github.com/EyeSeeTea/metadata-synchronization/pull/134/commits/d4eebaf7a6ac2897e547c59cf1986b291d022208

src/components/rules-creation-page/steps/SaveStep.jsx
SferaDev: See https://github.com/EyeSeeTea/metadata-synchronization/pull/134/commits/d4eebaf7a6ac2897e547c59cf1986b291d022208

src/components/rules-list-page/SyncRulesPage.jsx
SferaDev: See https://github.com/EyeSeeTea/metadata-synchronization/pull/134/commits/d4eebaf7a6ac2897e547c59cf1986b291d022208

PR #133, opened 2019-07-16
[Scheduler] Last executed property and ObjectsTable filters

### LINKED ISSUE #131: Add filters to sync-rules page for scheduling
- [x] Add filters to Sync Rules Page:
   - [x] Enabled/Disabled
   - [x] Last execution (Date)

SferaDev:
### :pushpin: References

* **Issue:** Closes #131 

### :memo: Implementation

- Update lastExecuted date prop of syncRules during synchronization

- Add new filters to the sync rules listing page

### :bookmark_tabs: Others

https://github.com/EyeSeeTea/d2-ui-components/pull/52


src/models/syncRule.ts
```diff
@@ -79,16 +84,25 @@ export default class SyncRule {
...
-            : data;
+        return {
+            ...data,
+            objects: _(data.objects)
+                .filter(rule =>
+                    targetInstanceFilter
+                        ? rule.builder.targetInstances.includes(targetInstanceFilter)
+                        : true
+                )
+                .filter(rule => (enabledFilter ? rule.enabled === enabledFilter : true))
```
tokland: No need to change anything, it's perfectly fine, but let me digress. In this pattern, to avoid the filtering when the guard condition is false, we could write: `.thru(rules => targetInstanceFilter ? rules.filter(...) : rules)`. Maybe someday we should simply create a mixin `_.filterIf(condition: boolean, predicate: (item: T) => boolean): T[]` :)

src/models/syncRule.ts
```diff
@@ -79,16 +84,25 @@ export default class SyncRule {
...
+                .filter(rule =>
+                    lastExecutedFilter
+                        ? moment(lastExecutedFilter).isSameOrBefore(rule.lastExecuted)
```
tokland: Really minor, but for rules without lastExecuted, this conditions returns true (`now.isSameOrBefore(undefined)` returns true, who knows why). Adding `rule.lastExecuted && ...` fixes it.

SferaDev: See https://github.com/EyeSeeTea/metadata-synchronization/pull/134/commits/8623619bcdad92a6b23fb04d68617e54497dd811

adrianq:
@SferaDev Not sure if we discussed this but can we add a column with the last executed date? It is a bit strange to filter by last executed but you cannot see this info on the table. What do you think?

adrianq:
My last comment has been solved on https://github.com/EyeSeeTea/metadata-synchronization/commit/85f60e77c76854b5785b4f0e94e0d691b2ed1dba

PR #138, opened 2019-07-29
Deleted objects

### LINKED ISSUE #100: [User Feedback] Sync deleted elements on destination
- [x] New option "Deleted Objects" in the landing page (in principle in third row)
- [x] Implement deleted objects section
  - [x] Single table with all pieces of metadata
  - [ ] Define deleting mapping for each piece of metadata
- [ ] New step in sync rule to include "deleted objects". Toggle per each piece of metadata-> "Synchronise deleted data elements" or multi-selector

SferaDev:
### :pushpin: References

* **Issue:** Closes #100 
* **Depends on:** https://github.com/EyeSeeTea/d2-ui-components/pull/57

### :memo: Implementation

- Add a new SyncPage that hosts all the deleted objects of the instance

- Before asking to delete the objects on destination, we query destination to see if they still exists

- We have re-used as much code as possible, everything works with the same Generic Sync Page that we use for programmatic metadata synchronizations.

### :art: Screenshots

image

- Deletion is not a fully working feature of DHIS2, for example it will fail if you are deleting an organisation that has children or if you are deleting an object that has been re-created with the same id after being deleted.

### DIFF
--- a/src/components/rules-creation-page/steps/SaveStep.jsx
+++ b/src/components/rules-creation-page/steps/SaveStep.jsx
@@ -53,7 +53,7 @@ const SaveStep = ({ d2, syncRule, classes, onCancel, snackbar }) => {
     };
 
     useEffect(() => {
-        getMetadata(d2, syncRule.metadataIds, "id,name").then(updateMetadata);
+        getMetadata(d2.Api.getApi().baseUrl, syncRule.metadataIds, "id,name").then(updateMetadata);
         getInstances(d2).then(setInstanceOptions);
     }, [d2, syncRule]);
 
--- a/src/components/rules-creation-page/steps/SaveStep.jsx
+++ b/src/components/rules-creation-page/steps/SaveStep.jsx
@@ -6,6 +6,7 @@ import { Button, LinearProgress, withStyles } from "@material-ui/core";
 import { ConfirmationDialog, withSnackbar } from "d2-ui-components";
 
 import { getInstances } from "./InstanceSelectionStep";
+import { getBaseUrl } from "../../../utils/d2";
 import { getMetadata } from "../../../utils/synchronization";
 import { getValidationMessages } from "../../../utils/validations";
 
@@ -53,7 +54,7 @@ const SaveStep = ({ d2, syncRule, classes, onCancel, snackbar }) => {
     };
 
     useEffect(() => {
-        getMetadata(d2.Api.getApi().baseUrl, syncRule.metadataIds, "id,name").then(updateMetadata);
+        getMetadata(getBaseUrl(d2), syncRule.metadataIds, "id,name").then(updateMetadata);
         getInstances(d2).then(setInstanceOptions);
     }, [d2, syncRule]);
 
--- a/src/components/sync-dialog/SyncDialog.jsx
+++ b/src/components/sync-dialog/SyncDialog.jsx
@@ -2,11 +2,10 @@ import React from "react";
 import i18n from "@dhis2/d2-i18n";
 import PropTypes from "prop-types";
 import _ from "lodash";
-import { ConfirmationDialog, MultiSelector, withLoading } from "d2-ui-components";
+import { ConfirmationDialog, MultiSelector } from "d2-ui-components";
 import DialogContent from "@material-ui/core/DialogContent";
 
 import Instance from "../../models/instance";
-import { startSynchronization } from "../../logic/synchronization";
 
 class SyncDialog extends React.Component {
     static propTypes = {
@@ -14,7 +13,7 @@ class SyncDialog extends React.Component {
         isOpen: PropTypes.bool.isRequired,
         metadataIds: PropTypes.array.isRequired,
         handleClose: PropTypes.func.isRequired,
-        loading: PropTypes.object.isRequired,
+        task: PropTypes.func.isRequired,
     };
 
     state = {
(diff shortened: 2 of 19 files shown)


src/models/deletedObjects.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    public get deletedAt(): Date {
+        return this.data.deletedAt;
+    }
+
+    public get deletedBy(): string {
+        return this.data.deletedBy;
+    }
+
+    public static build(data: any | undefined): DeletedObject {
```
tokland: `any` -> `DeletedObjectData` works?

src/utils/synchronization.ts
```diff
@@ -68,7 +70,8 @@ export async function getMetadata(
 
 export async function postMetadata(
     instance: Instance,
-    metadata: any
+    metadata: any,
```
tokland: A bit more descriptive: `any` -> `object`.

src/logic/delete.ts
```diff
@@ -0,0 +1,61 @@
+import i18n from "@dhis2/d2-i18n";
+
+import Instance from "../models/instance";
```
tokland: On Instances pages I get: "Cannot read property 'modelValidations' of undefined". That's because we need to use the d2-ui-components branch `feature/objects-table-improvements-10`?

src/components/sync-dialog/SyncDialog.jsx
```diff
@@ -2,19 +2,18 @@ import React from "react";
 import i18n from "@dhis2/d2-i18n";
```
tokland: On Sync error, the button [synchronize] remains disabled:

image

Btw, the error was that I kept a trailing "/" on the instance URL, maybe we should clean it up on save?

src/logic/metadata.ts
```diff
@@ -48,3 +49,22 @@ export async function listByIds(
...
+    originalParams?: Params
+): Promise<Pick<ModelCollection, "toArray" | "pager">> {
+    const { deletedObjects, pager } = (await axios.get(
+        d2.Api.getApi().baseUrl + "/deletedObjects",
+        {
+            withCredentials: true,
+            params: {
+                fields: ":all",
+                ...originalParams,
+                paging: false, // DHIS2 deletedObjects endpoint pager does not work properly
```
tokland: That's sad. It's always returning the first page, right? I checked, the problem is here, the pager has a `total=0` that messes things up:

https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-api/src/main/java/org/hisp/dhis/deletedobject/DeletedObjectQuery.java#L178

It's still in dev, I guess we should report it...

src/components/rules-creation-page/steps/SaveStep.jsx
```diff
@@ -53,7 +53,7 @@ const SaveStep = ({ d2, syncRule, classes, onCancel, snackbar }) => {
     };
 
     useEffect(() => {
-        getMetadata(d2, syncRule.metadataIds, "id,name").then(updateMetadata);
+        getMetadata(d2.Api.getApi().baseUrl, syncRule.metadataIds, "id,name").then(updateMetadata);
```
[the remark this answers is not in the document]
SferaDev: We are now calling getMetadata for destination instances. See https://github.com/EyeSeeTea/metadata-synchronization/blob/feature/deleted-objects/src/logic/delete.ts#L48

Chainging the effect a bit to make it more readable:

```
    useEffect(() => {
        const baseUrl = d2.Api.getApi().baseUrl;
        getMetadata(baseUrl, syncRule.metadataIds, "id,name").then(updateMetadata);
        getInstances(d2).then(setInstanceOptions);
    }, [d2, syncRule]);
```

src/logic/delete.ts
SferaDev: Yeah, done!

src/logic/metadata.ts
```diff
@@ -48,3 +49,22 @@ export async function listByIds(
 
     return { objects: paginatedData, pager: { page, total } };
 }
+
+export async function getDeletedObjects(
+    d2: D2,
+    originalParams?: Params
+): Promise<Pick<ModelCollection, "toArray" | "pager">> {
+    const { deletedObjects, pager } = (await axios.get(
+        d2.Api.getApi().baseUrl + "/deletedObjects",
```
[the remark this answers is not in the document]
SferaDev:
```
import { getBaseUrl } from "../utils/d2";
const baseUrl = getBaseUrl(d2);
```

src/logic/metadata.ts
SferaDev: https://jira.dhis2.org/browse/DHIS2-7381

src/components/synchronization-page/DeletedObjectsPage.jsx
```diff
@@ -0,0 +1,68 @@
+import React from "react";
```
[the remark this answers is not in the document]
SferaDev: Appears to happen with all metadata tables

image

I imagine it's 2 renders * 2 calls (with and without paging), the thing with deleted objects is that we are mocking the paging call to be non paged and all the requests are the same.

src/components/sync-dialog/SyncDialog.jsx
SferaDev:
```
    public setUrl(url: string): Instance {
        return new Instance({ ...this.data, url: url.replace(/\/+$/, "") });
    }
```

src/components/synchronization-page/DeletedObjectsPage.jsx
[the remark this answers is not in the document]
tokland: Not sure if this is PR, but eventually we should try to clean this.

src/components/rules-creation-page/steps/SaveStep.jsx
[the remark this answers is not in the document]
tokland: Oh, I see. Looks good with that helper.

src/components/synchronization-page/DeletedObjectsPage.jsx
```diff
@@ -20,46 +20,49 @@ class DeletedObjectsPage extends React.Component {
...
+                },
+            }),
+        },
+    ];
 
     render() {
         const { d2 } = this.props;
 
         const title = i18n.t("Deleted Objects Synchronization");
+        const buttonLabel = <DeleteIcon />;
```
tokland: AFIK, `DeleteIcon` is always the same, but `<DeleteIcon />` instances a new component each time.

SferaDev: Oh! Since we ask for a component in ObjectsTable moving it outside the render method should fix it then.

PR #139, opened 2019-07-29
Force selection column for ObjectTable in sync pages

tokland:
### :pushpin: References

* **PR:** https://github.com/EyeSeeTea/d2-ui-components/pull/57

### :memo: Implementation

- Use new prop to force the column showing.


adrianq:
This PR already went to development in another PR. It looks like there is no need to merge it

PR #142, opened 2019-09-03
Allow selection a whole OU subtree

### LINKED ISSUE #141: Add an action per row (right-click or clicking on the 3-dots at the end of the row) to select all children of one or many organisation units

SferaDev:
### :pushpin: References

* **Issue:** Closes #141 

### :memo: Implementation

- Update dependencies to fix vulnerability
- Fix bug that crashed app when ordering by OU level
- Add a new way to select a whole sub-tree on OrganisationUnit's Synchronization Page

- ObjectsTable does not allow to programmatically select items from outside the table
- The only work-around possible is to use initialSelection and ask the table to re-render
- Due to this limitation, you might lose part of the selection when using this action (selection in other pages)
- This limitation could be overcome if the action is not ```multiple``` as you would be forced to have selection only in the current page

### :art: Screenshots 

image

image

image


i18n/en.pot
```diff
@@ -428,6 +428,9 @@ msgstr ""
 msgid "Indicators Synchronization"
 msgstr ""
 
+msgid "Select children subtree"
+msgstr ""
```
tokland: Isn't a bit misleading, as it also selects the parent org unit?

src/components/synchronization-page/GenericSynchronizationPage.jsx
```diff
@@ -39,6 +39,8 @@ class GenericSynchronizationPage extends React.Component {
     };
 
     changeSelection = metadataIds => {
+        const { clearSelection } = this.props;
+        if (metadataIds.length === 0) clearSelection();
```
tokland: It should be added to `propTypes`. However, I an not sure about this prop, couldn't we use a more generic `onSelectionChange`? This is related with the comment in the PR: 

> Due to this limitation, you might lose part of the selection when using this action (selection in other pages)

MetadataTable has `notifyNewSelection`, can't we use it also in `GenericSynchronizationPage` so `OrganisationUnitPage` knows the previous selection on `selectChildren`?

src/components/synchronization-page/OrganisationUnitPage.jsx
```diff
@@ -1,25 +1,70 @@
...
+    };
+
     models = [OrganisationUnitModel, OrganisationUnitGroupModel, OrganisationUnitGroupSetModel];
 
+    clearChildren = () => this.setState({ children: [] });
+
+    selectChildren = async selectedOUs => {
+        const { d2 } = this.props;
+        const children = _.flatten(
+            await Promise.all(selectedOUs.map(selectedOU => getOrgUnitSubtree(d2, selectedOU.id)))
```
tokland: Risky unbounded parallel requests, specially when you say that `getOrgUnitSubtree` is a "very dangerous operation"  that tends to crash the server :). I keep forgetting if we have a promise pool in this project.

src/components/synchronization-page/OrganisationUnitPage.jsx
```diff
@@ -1,25 +1,70 @@
...
 import GenericSynchronizationPage from "./GenericSynchronizationPage";
 
 export default class OrganisationUnitPage extends React.Component {
     static propTypes = {
         d2: PropTypes.object.isRequired,
     };
 
+    state = {
+        children: [],
+        metadataTableKey: Math.random(),
```
tokland: > The only work-around possible is to use initialSelection and ask the table to re-render

Yep, at some point we should probably add some prop to ObjectsTable (`selection: string[]`) so it also works in controlled mode. For now, this workaround should do the trick, the only problem is that it makes some extra requests that could be avoided.

i18n/en.pot
SferaDev: Maybe "Select with children subtree"

src/components/synchronization-page/GenericSynchronizationPage.jsx
SferaDev: Yeah, this part is really "tricky".

If we add notifyNewSelection in "OrganisationUnitPage" we will override the one in "GenericSynchronizationPage". My inner java would love to have a subscription-like method to attach to selection changes without taking care of how many observers it has.

I would also love to have a method that programmatically selects a collection of ids from outside of the ObjectsTable but we were trying to have a quick fix for this to ship as soon as possible.

I would propose for the next "big" app project to re-work ObjectsTable and allow uncontrolled behavior for the objects and selection collections.

Meanwhile, to make it more generic I will add another onSelectionChange that does not override the other one and add it to the prop-types.

src/components/synchronization-page/OrganisationUnitPage.jsx
SferaDev: Yeah, initially the action was not multiple. I will replace the ```promise.all``` with a ```for of async```.

I faced the crashing of the server several times with that request on 2.28 WHO servers a year ago. We had to completely deprecate that call at the time but now on 2.30 appears to work fine, even though it is still a very demanding request. I prefer to have the warning in the code, just in case!

adrianq:
@SferaDev The feature seems to be working fine. I cannot even reproduce the error regarding losing part of the selection. However, there seems to be a major error here. Steps to reproduce it:

1. Go to OU sync
1. Choose in OU Level filter -> 4
1. Select with children subtree the very first one
1. Click on Sync and you get:
image

PR #154, opened 2019-10-10
Add new metadata types for programs

### LINKED ISSUE #149: Add new metadata types
- [ ] Program Indicator
- [ ] Program Indicator Groups
- [ ] Program Rules
- [ ] Program Rules Variables

SferaDev:
### :pushpin: References

* **Issue:** Closes #149 
* I am still testing this PR but uploading for @adrianq to also begin with the testing

### :memo: Implementation

- Program Indicators: Followed the same as per Indicators. The main difference is ```analyticsPeriodBoundaries``` but the object appears to be bundled inside.

- Program Indicator Groups: Followed the same as per Indicator Groups.

- Program Rules: Has a very strong relationship with ``program`` and I find dangerous synchronizing this metadata without controlling the element exists on remote.

- Program Rule variable: Has a very strong relationship with ``program``. ``programStage`` and ``dataElement``. This appears to be a dangerous uncontrolled operation.

For future versions of metadata-sync (or data-sync) we could try to automate the verification of some of these relationships in destination.

### :art: Screenshots

image

- There are two WHO instances available for test, ask for credentials if needed.

- WHO does not currently use Program Indicator Groups but we want to push them so they begin using them for this synchronization purposes.

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2019-09-17T10:45:09.309Z\n"
-"PO-Revision-Date: 2019-09-17T10:45:09.309Z\n"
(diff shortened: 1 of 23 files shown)


adrianq:
@SferaDev I am merging this to development but I will wait for your confirmation that you have fully tested this feature before merging it to master

PR #155, opened 2019-10-10
Sharing settings and administrative user role

### LINKED ISSUE #148: Sharing settings for sync rules
- [x] Add default to Sync Rules wizard
- [x] Add context action to edit sync rules

Schema of sharing settings:

*Default*

```
"publicAccess": "rw------", // No public access
"userGroupAccesses": []
```

*Example*

```
"publicAccess": "--------", // No public access
"userGroupAccesses": [{
	"access": "r-r-----",
	"id": "QhhxtdJ8HL7" // NHWA Admins (Department in WHO)
}, {
	"access": "rwrw----",
	"id": "Urx6COfeLkm" // WIDP IT Team
}]
```

References:

- https://github.com/dhis2/maintenance-app/blob/37714d007a829e9fd79412c344144a132c4f647f/src/EditModel/option-set/OptionManagement.component.js

- https://github.com/EyeSeeTea/dataset-configuration/blob/development/src/DataSets/DataSets.component.js#L494

- https://github.com/EyeSeeTea/d2-ui/blob/master/src/sharing/SharingDialog.component.js

For later:

- Encryption for sharing settings -> See #181

SferaDev:
### :pushpin: References

* **Issue:** Closes #148 

### :memo: Implementation

- Visual component for Sharing settings
  - Add ``downshift`` dependency
  - Import Sharing component from d2-ui
  - Remove old ``d2.i18n`` code, ``recompose`` library and logic coupling inside the components
  - Expose the logic methods as props in the components to make them agnostic (allowing dataStore support)
- Administrative user roles
  - On application load it checks if the user role exists and creates it if negative
  - For creation of sync rules and/or instances we verify user has the administrative role
- Sharing settings and user access
  - For ``SyncRule.list`` we filter the results with the ones the user has access (read)
  - For contextual actions (edit, delete, sharing settings, toggle enabled) we verify user has access (write)
  - We support public access rules (both read and/or write)
  - We allow administrative users to always have read and write permissions (as DHIS2 does)

### :art: Screenshots

image

image

- The administrative user role is called "METADATA_SYNC_ADMINISTRATOR"
- The original issue contemplated two user roles (configurators/executors) but this is superseded by one single user role (configurator) + sharing settings (executors only are meant to view/execute the rules they have access to)

### :bookmark_tabs: Others

- When we have ready our logic library I would like to revisit this part and add more robust methods (using d2-api).
- Version 2.31 adds support for dataStore sharingSettings natively in the /sharings endpoint, I would like to add that extra validations but does not make sense doing it until we have our communication library

### DIFF
--- a/src/components/app/App.jsx
+++ b/src/components/app/App.jsx
@@ -13,7 +13,7 @@ import Share from "../share/Share";
 import Instance from "../../models/instance";
 import { muiTheme } from "./themes/dhis2.theme";
 import muiThemeLegacy from "./themes/dhis2-legacy.theme";
-import { initializeAppRoles } from "../../utils/permissions";
+import { initializeAppRoles, isUserAdmin } from "../../utils/permissions";
 import "./App.css";
 
 const generateClassName = createGenerateClassName({
@@ -27,7 +27,11 @@ class App extends Component {
         appConfig: PropTypes.object.isRequired,
     };
 
-    componentDidMount() {
+    state = {
+        isUserAdmin: false
+    }
+
+    async componentDidMount() {
         const { d2, appConfig } = this.props;
         const appKey = _(this.props.appConfig).get("appKey");
 
@@ -44,10 +48,14 @@ class App extends Component {
         }
 
         initializeAppRoles(d2.Api.getApi().baseUrl);
+
+        const isAdmin = await isUserAdmin(d2);
+        this.setState({ isAdmin})
     }
 
     render() {
         const { d2, appConfig } = this.props;
+        const { isAdmin } = this.state;
         const showShareButton = _(appConfig).get("appearance.showShareButton") || false;
         const showHeader = !process.env.REACT_APP_CYPRESS;
 
@@ -63,7 +71,7 @@ class App extends Component {
                                     )}
 
                                     <div id="app" className="content">
-                                        <Root d2={d2} />
+                                        <Root d2={d2} isAdmin={isAdmin} />
                                     </div>
 
                                     <Share visible={showShareButton} />
--- a/src/components/app/App.jsx
+++ b/src/components/app/App.jsx
@@ -28,8 +28,8 @@ class App extends Component {
     };
 
     state = {
-        isUserAdmin: false
-    }
+        isUserAdmin: false,
+    };
 
     async componentDidMount() {
         const { d2, appConfig } = this.props;
@@ -50,7 +50,7 @@ class App extends Component {
         initializeAppRoles(d2.Api.getApi().baseUrl);
 
         const isAdmin = await isUserAdmin(d2);
-        this.setState({ isAdmin})
+        this.setState({ isAdmin });
     }
 
     render() {
--- a/src/components/app/App.jsx
+++ b/src/components/app/App.jsx
@@ -13,6 +13,7 @@ import Share from "../share/Share";
 import Instance from "../../models/instance";
 import { muiTheme } from "./themes/dhis2.theme";
 import muiThemeLegacy from "./themes/dhis2-legacy.theme";
+import { initializeAppRoles } from "../../utils/permissions";
 import "./App.css";
 
 const generateClassName = createGenerateClassName({
@@ -41,6 +42,8 @@ class App extends Component {
         if (appConfig && appConfig.encryptionKey) {
             Instance.setEncryptionKey(appConfig.encryptionKey);
         }
+
+        initializeAppRoles(d2.Api.getApi().baseUrl);
     }
 
     render() {
--- a/src/components/app/App.jsx
+++ b/src/components/app/App.jsx
@@ -13,7 +13,7 @@ import Share from "../share/Share";
 import Instance from "../../models/instance";
 import { muiTheme } from "./themes/dhis2.theme";
 import muiThemeLegacy from "./themes/dhis2-legacy.theme";
-import { initializeAppRoles, isUserAdmin } from "../../utils/permissions";
+import { initializeAppRoles } from "../../utils/permissions";
 import "./App.css";
 
 const generateClassName = createGenerateClassName({
@@ -25,10 +25,6 @@ class App extends Component {
     static propTypes = {
         d2: PropTypes.object.isRequired,
         appConfig: PropTypes.object.isRequired,
-    };
-
-    state = {
-        isUserAdmin: false,
     };
 
     async componentDidMount() {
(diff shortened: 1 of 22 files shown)


SferaDev:
Before merging I would like to add a validation in the router for create/edit pages. Right now going directly to the url allows anyone to creatw/edit. Will do tomorrow morning.

adrianq:
@SferaDev Still testing but a few comments:
- The two roles were needed as per the client requirements. As we are now in a rush because we only have a few hours for testing and discussions I have created a new card to take care of this https://github.com/EyeSeeTea/metadata-synchronization/issues/156 We will discuss together when we take care of this task.
- I am using your dev server for testing. I have created a rule (testing again) with the user EST, made the rule private (no public access, no user and no user group), created another user EST2. EST2 can see the rule and as far as I know this should not be the case. Am I missing something?

adrianq:
It is an admin of the app but not a global admin. I would say superusers (global admin) can see everything but for admins of the app we should still apply sharing settings.

src/components/app/Root.jsx
```diff
@@ -20,58 +20,65 @@ class Root extends React.Component {
     };
 
     render() {
-        const { d2 } = this.props;
+        const { isAdmin } = this.props;
```
tokland: `isAdmin` missing from propTypes

src/components/rules-list-page/SyncRulesPage.jsx
```diff
@@ -201,6 +206,37 @@ class SyncRulesPage extends React.Component {
...
+    closeSharingSettings = () => {
+        this.setState({ sharingSettingsObject: null });
+    };
+
+    verifyUserHasPermissions = (d2, rule) => {
+        const { isAdmin } = this.props;
+        const { id: userId, userGroups } = this.state.userInfo;
+        const { publicAccess = "--------", userAccesses = [], userGroupAccesses = [] } = rule;
+
+        return (
```
tokland: That's a lot of logic for a component, maybe method in syncRule?

src/components/rules-list-page/SyncRulesPage.jsx
```diff
@@ -244,6 +291,27 @@ class SyncRulesPage extends React.Component {
 
     changeLastExecutedFilter = moment => this.setState({ lastExecutedFilter: moment });
 
+    onSearchRequest = key =>
+        this.props.d2.Api.getApi()
+            .get("sharing/search", { key })
+            .then(searchResult => searchResult);
```
tokland: I guess you were planning to do some transformation here?

src/components/sharing-dialog/Access.jsx
```diff
@@ -0,0 +1,165 @@
...
+                <div>{primaryText}</div>
+                <div style={{ color: "#818181", paddingTop: 4 }}>{secondaryText || " "}</div>
+            </div>
+            <PermissionPicker
+                access={access}
+                accessOptions={accessOptions}
+                onChange={onChange}
+                disabled={disabled}
+            />
+            <IconButton disabled={!onRemove} onClick={onRemove || (() => {})}>
```
tokland: [minor] Use defaultProps for this defaults. Ok, I see it comes from the original code, no problem, we cannot change everything.

adrianq:
Something else, Edit feature doesn't seem to be working fine. Link visibility seems to be working properly based on the sharing settings but when you click on the link nothing happens

src/components/sharing-dialog/AutoComplete.jsx
```diff
@@ -0,0 +1,156 @@
...
+    container: {
+        flexGrow: 1,
+        position: "relative",
+    },
+    inputRoot: {
+        flexWrap: "wrap",
+    },
+});
+
+let popperNode;
```
tokland: Can we use an instance variable within Autocomplete?

src/utils/permissions.ts
```diff
@@ -0,0 +1,68 @@
...
+    return userRoles
+        .toArray()
+        .find((role: any) => role.name === AppRoles.METADATA_SYNC_ADMINISTRATOR);
+});
+
+export const getUserInfo = memoize(
+    async (
+        d2: D2
+    ): Promise<{
+        userGroups: any[];
```
tokland: As you hinted in the PR comment, those `any` should be real interfaces when `d2-api` gets integrated.

src/utils/permissions.ts
```diff
@@ -0,0 +1,68 @@
...
+            withCredentials: true,
+            params: {
+                userRoles: true,
+                filter: `name:eq:${role}`,
+                fields: "id",
+            },
+        })).data as { userRoles?: { id: string }[] };
+
+        if (!userRoles || userRoles.length === 0) {
+            await axios.post(
```
tokland: Some validation maybe handy, nothing fancy, just checking status for example.

src/components/rules-list-page/SyncRulesPage.jsx
```diff
@@ -201,6 +206,37 @@ class SyncRulesPage extends React.Component {
...
+    };
+
+    closeSharingSettings = () => {
+        this.setState({ sharingSettingsObject: null });
+    };
+
+    verifyUserHasPermissions = (d2, rule) => {
+        const { isAdmin } = this.props;
+        const { id: userId, userGroups } = this.state.userInfo;
+        const { publicAccess = "--------", userAccesses = [], userGroupAccesses = [] } = rule;
```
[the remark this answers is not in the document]
SferaDev: Yeah since the sharing endpoint already retieves both user/userGroups I included both. The minimum requirements was to include userGroups only but having extra functionality is always welcomed.

src/components/app/App.jsx
```diff
@@ -41,10 +46,16 @@ class App extends Component {
         if (appConfig && appConfig.encryptionKey) {
             Instance.setEncryptionKey(appConfig.encryptionKey);
         }
+
+        initializeAppRoles(d2.Api.getApi().baseUrl);
```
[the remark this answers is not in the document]
SferaDev: I guess it does no harm to wait for it, will do.

src/components/app/Root.jsx
SferaDev: Yeah I originally had it set in ```state``` everywhere and moved to props in the last commit. Will add to propTypes in the next commit.

src/components/rules-list-page/SyncRulesPage.jsx
```diff
@@ -201,6 +206,37 @@ class SyncRulesPage extends React.Component {
...
+
+        return (
+            isAdmin ||
+            publicAccess.substring(0, 2).includes("w") ||
```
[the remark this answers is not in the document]
SferaDev: Sounds good, a helper method could come in handy (in SyncRule class)

src/components/rules-list-page/SyncRulesPage.jsx
SferaDev: Sounds good

src/components/rules-list-page/SyncRulesPage.jsx
SferaDev: Yeah, will clean it up. Thanks!

src/components/sharing-dialog/Access.jsx
SferaDev: I tried to only touch ``breaking`` changes, but since it's a simple change I will include it in next commit.

src/utils/permissions.ts
SferaDev: This will only happen on first load of the application or if someone deleted the roles manually.

I wanted to control the scenario where the app has just been opened for the first time and the user does not have the authority to import metadata (which will likely fail with 200-ERROR).

PR #186, opened 2019-10-18
Fix/model centralization

### LINKED ISSUE #184: Fix model centralization

xurxodev:
### :pushpin: References

* **Issue:** Closes #184 

### :memo: Implementation
After some testing, the problem seems to be the use of keyBy Lodash method in d2ModelFactory.
For some reason I don't understand, after building with WebPack the classes dictionary in the production environment only the last model in array ProgramRuleVariableModel, and in the development environment had all.

- Solution:
I have changed the strategy to centralize metadataModels array without to use keyBy Lodash method. I have restored the old approach in d2ModelFactory as @SferaDev suggested and I have added a new export metadataModels array here based on classes dictionary. 
I have removed metadataModels array from D2Model and all places where was used now the import has been changed to metadataModels in d2ModelFactory.

* Trace
I have added a console.log when d2ModelFactory return defaultModel

* Testing
I have added a new unit test for d2ModelFactory but this jest test works on the development and Travis even with the error because jest tests are executed in the prebuild phase.
I think the unique way the test is failing in Travis when a similar problem occurs is creating an integration test.
The problem is the time needed to invest to create this type of test because it is necessary to create a test environment data totally controlled by the tests so as not to depend on data from a specific server of dhis2 since they would be very brittle tests.

**UseCase 1**: try to synchronize dataElementGroups then the final summary should to show that Datalements has been imported also
**UseCase 2**: In the metadata sync page on metadata type dropdown should appear all metadata models.
**UseCase 3**: In the metadata step of metadata sync rules wizard, on metadata type dropdown should appear all metadata models.

### DIFF
--- a/src/models/d2ModelFactory.ts
+++ b/src/models/d2ModelFactory.ts
@@ -1,8 +1,43 @@
-import { D2Model, defaultModel, metadataModels } from "./d2Model";
+import {
+    D2Model, defaultModel, DataElementModel,
+    DataElementGroupModel,
+    DataElementGroupSetModel,
+    IndicatorModel,
+    IndicatorGroupModel,
+    IndicatorGroupSetModel,
+    OrganisationUnitModel,
+    OrganisationUnitGroupModel,
+    OrganisationUnitGroupSetModel,
+    ValidationRuleModel,
+    ValidationRuleGroupModel,
+    ProgramIndicatorModel,
+    ProgramIndicatorGroupModel,
+    ProgramRuleModel,
+    ProgramRuleVariableModel
+} from "./d2Model";
+
 import { D2 } from "../types/d2";
 import _ from "lodash";
 
-const classes: { [modelName: string]: typeof D2Model } = _.keyBy(metadataModels, o => o.name);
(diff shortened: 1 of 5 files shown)


src/models/d2ModelFactory.ts
```diff
@@ -12,5 +48,13 @@ const classes: { [modelName: string]: typeof D2Model } = _.keyBy(metadataModels,
 export function d2ModelFactory(d2: D2, d2ModelName: string): typeof D2Model {
     const modelName = d2.models[d2ModelName].name;
     const className = modelName.charAt(0).toUpperCase() + modelName.slice(1) + "Model";
-    return classes[className] || defaultModel(modelName);
+
+    let model = classes[className];
```
SferaDev: We use to use ``const`` for almost everything. I believe we could here keep the ``return classes[className] || defaultModel(modelName);`` and just add before the return an ``if (!classes[class ame])``.

adrianq:
Works like a charm! I have just changed the log to debug and now we are printing every single model requested to the factory. Not sure if it is too much... Kudos for @SferaDev that pointed us to the right bit of code!

PR #191, opened 2019-10-23
Update to v4 material-ui

SferaDev:
### :memo: Implementation

- Update to material v4 as skeleton-app does (to make sure we are in the same major version everywhere)

- Remove unnecessary dependencies

- Adjust breaking change in Grid for the landing-page


SferaDev:
@tokland The visual bugs that we are facing are because the material-ui theme that we apply on the app never gets applied in ``d2-ui-components``.

I have tested the following:

- FAB in ObjectsTable now has color ``indigo`` as background
- Copy and use ``ListActionBar.component.js`` from the ObjectsTable inside the app
- FAB is correctly styled with our custom ``primary`` color

Do you know what could be causing this issue?

tokland:
@SferaDev We now have different versions of material-ui: v3 in d2-ui-comp, v4 in the app, so they are not using the same styling context. Either we update d2-ui-components to v4 or I am afraid we'll have to wrap every exposed component with its v3 MuiThemeProvider.

tokland:
@SferaDev We've decided with @adrianq to tackle the v4 migration very soon -> https://github.com/EyeSeeTea/d2-ui-components/issues/74

tokland:
This PR updates v3->v4 and v0->v4:

https://github.com/EyeSeeTea/d2-ui-components/pull/83

adrianq:
Guys, I am merging this now. In general, look&feel seems to be quite similar to the original one but let's keep an eye because there could be some minor glitches.

The only thing I have noticed is that the snack bar is not located anymore at the bottom of the page but there seems to be a bit of bottom margin. @SferaDev I have created this issue https://github.com/EyeSeeTea/metadata-synchronization/issues/203 and assigned it to you. It would be great if you can have a look at it on Tuesday next week.

SferaDev:
> Guys, I am merging this now. In general, look&feel seems to be quite similar to the original one but let's keep an eye because there could be some minor glitches.
> 
> The only thing I have noticed is that the snack bar is not located anymore at the bottom of the page but there seems to be a bit of bottom margin. @SferaDev I have created this issue #203 and assigned it to you. It would be great if you can have a look at it on Tuesday next week.

Sure thing

PR #192, opened 2019-10-24
Allow sync rule owners

### LINKED ISSUE #173: Allows private sharings
A user should be always able to access its own sync rules even if they are set as private. For this to happen we need to store the owner of a sync rule in the data store. This attempts to mimic DHIS2 behaviour in terms of sharing

SferaDev:
### :pushpin: References

* **Issue:** Closes #173

### :memo: Implementation

- Add code, user, created, lastUpdated, lastUpdatedBy properties

- Allow introducing code in the sync rules wizard

- Allow read/write access to the user who created a given sync rule.

### :art: Screenshots

image

image

image

- Remember that even if the user created the sync rule, if he/she does not have EXECUTION authority will be unable to execute the rule.

- To maintain backwards compatibility we are checking if the ``user`` property exist even if it is mandatory to be non-null as per the type. @tokland Do you have a better approach on this?

### DIFF
--- a/src/types/synchronization.d.ts
+++ b/src/types/synchronization.d.ts
@@ -1,3 +1,5 @@
+import { Ref } from "d2-api";
+
 import { MetadataImportResponse, MetadataImportStats, MetadataImportParams } from "./d2";
 import SyncReport from "../models/syncReport";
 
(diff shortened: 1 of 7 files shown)


src/types/synchronization.d.ts
```diff
@@ -58,15 +60,24 @@ export interface SynchronizationState {
     done?: boolean;
 }
 
+interface NamedRef extends Ref {
+    name: string;
+}
```
SferaDev: @tokland I believe I have seen this type of references somewhere in the API. I have pending to check if we already have a type in ``d2-api`` for this.

PR #195, opened 2019-10-25
Feature/selection of org units

### LINKED ISSUE #169: Selection of Org Units
For this phase, we will just use the Org Unit Tree selector component. To take care of the aggregation per org unit (see #164) we will have to modify the component

xurxodev:
### :pushpin: References

* **Issue:** Closes #169 

### :memo: Implementation

- I have added a new function to retrieve curren user org units according to a talk with @tokland 
- I have created a type and dataParams in SyncRule 
- I have created an optional parameter to create a method of SyncRule model assigning metadata by default to maintain current behaviour.
- I have modified validations in SyncRule because are differents for data or metadata SyncRule
- I have added new tests to verify data or metadata validations 

According to talk with @tokland levels and listParams props of OrgUnitsSelector are not necessary for this project. 

### :art: Screenshots

data-synchronization
data-synchronization-rule

**Use case 1**:  for on-demand data sync dialog in the organisationUnitStep to click on next if does not exists selected org units should show an error message
**Use case 2**:  for on-demand data sync dialog in the organisationUnitStep to click on next if exists selected org units should navigate to next step
**Use case 3**:  to create data sync rule in the organisationUnitStep to click on next if does not exists selected org units should show an error message
**Use case 4**:   to create data sync rule in the organisationUnitStep to click on next if exists selected org units should navigate to next step
**Use case 5**:   data sync rule should save in data store assigned type field to data
**Use case 6**:   metadata sync rule should save in data store assigned type field to metadata

### :bookmark_tabs: Others

- I have modified OrgUnitSelector in d2-ui-components to add a new prop to assign from outside if the shadow is visible:
PR: https://github.com/EyeSeeTea/d2-ui-components/pull/72

[Selection of Org Units](https://app.gitkraken.com/glo/board/XaBKjXlqfAAPh4Tg/card/XaBssnlqfAAPh4qY)


adrianq:
It works properly and data is stored in the data store. Something I couldn't test is the new prop to hide the shadow. Original PR https://github.com/EyeSeeTea/d2-ui-components/pull/72 was finally merged in https://github.com/EyeSeeTea/d2-ui-components/pull/83 that update material-ui to v4. We will have to double-check if it is working as expected once we update this repo to use the new d2-ui-comp lib.

PR #199, opened 2019-11-01
[DataTable] New component written in TS

SferaDev:
### :pushpin: References

* **DO NOT MERGE:** This has been developed here but meant to be moved to ``d2-ui-components`` once we have material v4 there.

* **Related issue:** https://github.com/EyeSeeTea/d2-ui-components/issues/68

#### File arrangement and structure

The granularity to divide components in different files is ``medium`` sized meaning only conceptually equivalent elements have been grouped in common files.

- utils
    - filtering: Utilities to allow search filters (this is only used in ObjectsTable and could be potentially moved there).
    - formatting: "Smart formatter" that allows correctly printing an element if it's an array of elements, d2-like object that provides displayName, name or id, date or moment or url. If nothing matches defaults to the element provided.
    - selection: Set of utility methods used to handle the selection of rows and actions. Includes the helper methods to create table notifications (select in all pages).
    - sorting: Set of utility methods to perform the sorting (this should be revisited very soon to improve the algorithm and allow automated sorting of d2-like elements).
- ActionButton: Floating action button used in ObjectsTable (refactored into TS from d2-ui-components).
- ContextualMenu: Popover menu printed when an action is requested.
- DataTable: Main component code.
- DataTableBody: Component similar to ``tbody``.
- DataTableHeader: Component similar to ``thead``.
- DataTableNotifications: Component refactored from d2-ui-components to show messages/warning/error on-demand (for example, select in multiple pages)
- DataTablePagination: Wrapper of the material-ui TablePagination.
- DetailsBox: Component refactored from d2-ui-components to show a details box.
- ObjectsTable: DataTable wrapper that adds 1) searchbox, 2) details box and 3) action button.
- types: Type definition file for generic types (prop types are in each component file)

#### Contextual actions

Checkboxes are hidden if no multiple action has been defined or ``forceSelectionColumn`` is set to false. Similarly if ``forceSelectionColumn`` is set to true checkboxes are always shown even though there are no multiple actions.

image

Icons are optional by default and expects a ``ReactNode``. In ObjectsTable the details icon is injected if no icon is provided. Similarly the details onClick action has injected the method to open the panel but still calls the parent onClick receiver.

image

Note that in the old ObjectsTable/DataTable we returned either an object or an array of objects for the onClick, now we always return an array of objects and it's safe to assume that for single actions the row will be located at array[0]. The code will never call the onClick if there are no selected rows, however I could not find a proper way to type that behavior.

#### Controlled/uncontrolled behavior

This is still a work-in-progress, the behavior in this pull request should be considered a first option. Right now we expose three optional props that act as uncontrolled props: ``pagination``, ``selection`` and ``sorting``.

The use cases for this uncontrolled behavior are the following:

- Pagination: Most common scenario. Allow pagination to next/prev pages not yet provided to the table, as in server-side pagination.

- Selection: Used in apps like metadata-sync. Allow selecting elements programmatically due to events not available inside the table.

- Sorting: Probably the least used scenario. Sort by a custom comparator (priority sorting, nested sorting, etc.). This could've been implemented with just a comparator func but for consistency I applied the same pattern as before.

How it works? If you pass the ``pagination``, ``selection`` or ``sorting`` props the table ignores the auto-computed value and just uses the provided one. However the internals still work and the events (change page, change selection, change sorting) are notified to parent through a ``onChange`` property method.

If only a partial-uncontrolled behavior is desired we also offer ``initialPagination``, ``initialSelection`` and ``initialSorting`` to preload the state instead of controlling it from the parent.

#### Positioning

Since we want to make the DataTable as customizable as possible we have added two props that allows inserting components either in the Toolbar (where the Pagination is located) and next to the DataTable itself (DetailsBox for example).

Both properties expect either a single ReactNode or a collection of them. Since they are rendered inside a ``flex`` layout, you can position them anywhere in the screen with the ``order`` property.

Allowing the application to do things like the following:

image

image

#### Table notifications

Previously we only included notifications related to the selection. We now allow to provide external customizable notifications.

image

#### Selection in all pages

The behavior is the same as we had in the original ObjectsTable, however now we do not perform two calls to a listing method with paging false as we previously did.

Now we control it through a prop called ``idsForSelectInAllPages`` (could be easily renamed) that has three different states:

- undefined: Default behavior. Auto-computed select in all pages according to the ``objects`` passed down as rows.
- string[]: Array of ``id`` strings that represent exisiting (or plausible) rows. We do not verify the objects are in rows as they could be in paginated rows not available to the table.
- null: Disables the select in all pages behavior, the message to select everywhere never appears. However if there are selected items in other pages or hidden trough filtering the other-pages-selection still appears.

image

- Beware this is a huge refactor, here be dragons!

### :bookmark_tabs: Things to revisit in future PRs

- Implement the filter columns feature and allow a way to force show/hide. Right now the icon is visible to keep in mind the TODO. Feel free to hide under an always false condition until ready. The same button could have multiple external actions (download as csv, etc) and show a wheel instead. To be discussed as-per the requirements of Data Sync and Samaritan's.

- Select in all pages: I just copied the same code we had in the previous ObjectsTable, this could be probably improved.

- Default sorting for decimal numbers does not work properly

- We are not doing any special memoization or optimization, as suggested by the react core team, it's better to create the components first and do a performance analysis and only tune the components that re-render the most or take a lot of time to re-render.

- Table notifications: Remove ``newSelection`` prop and properly handle ``onClick`` action.

- Either in d2-ui-components or d2-api we could create both a DataTable and ObjectsTable wrapper that manages the uncontrolled pagination and updates the sorting with server side object loading.

- Search box does not allow customizing ``hintText`` and ``initialValue`` in d2-ui-components. We should add the prop there since we pass the customization here. Probably refactoring it into TS altogether. Keep in mind that this is still material v0.


src/components/data-table/DataTableHeader.tsx
```diff
@@ -0,0 +1,113 @@
...
+    const { orderBy, order } = sorting;
+
+    const createSortHandler = (property: string) => (_event: React.MouseEvent<unknown>) => {
+        const isDesc = orderBy === property && order === "desc";
+        onChange({ orderBy: property, order: isDesc ? "asc" : "desc" });
+    };
+
+    return (
+        <TableHead>
+            <TableRow style={{ borderBottom: "3px solid #E0E0E0" }}>
```
SferaDev: Remove redundant style

src/components/data-table/ContextualMenu.tsx
```diff
@@ -0,0 +1,76 @@
...
+export interface ContextualMenuProps {
+    isOpen: boolean;
+    rows: TableObject[];
+    positionLeft: number;
+    positionTop: number;
+    closeContextMenu(): void;
+    actions: TableAction[];
+}
+
+//TODO: Icon
```
SferaDev: Implement custom icon

src/components/data-table/ContextualMenu.tsx
```diff
@@ -0,0 +1,76 @@
...
+            {actions.map(action => (
+                <MenuItem
+                    className={classes.item}
+                    key={action.name}
+                    onClick={() => {
+                        if (action.onClick) action.onClick(rows);
+                        closeContextMenu();
+                    }}
+                >
+                    <DetailsIcon className={classes.icon} fontSize="small" />
```
SferaDev: Details Icon should be "injected" in Objects Table. Default should not have any icon.

src/components/data-table/ContextualMenu.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+                    <Typography className={classes.text} variant="inherit" noWrap>
```
SferaDev: Variant inherit, remove?

src/components/data-table/ContextualMenu.tsx
```diff
@@ -0,0 +1,76 @@
...
+    positionLeft: number;
+    positionTop: number;
+    closeContextMenu(): void;
+    actions: TableAction[];
+}
+
+//TODO: Icon
+export function ContextualMenu(props: ContextualMenuProps) {
+    const classes = useStyles();
+    //@ts-ignore
```
SferaDev: ContextualMenuProps are already implemented, remove TS ignore

src/components/data-table/DataTableBody.tsx
```diff
@@ -0,0 +1,120 @@
...
+    } = props;
+    const classes = useStyles();
+    const isSelected = (name: string) => selected.indexOf(name) !== -1;
+
+    const contextualAction = (row: TableObject, event: React.MouseEvent<unknown>) => {
+        event.stopPropagation();
+        openContextualMenu(row, event.clientY, event.clientX);
+    };
+
+    const handleClick = (event: React.MouseEvent<unknown>, row: any) => {
```
SferaDev: Instead of two different castings we should have two methods one for row and another one for checkbox

src/components/data-table/DataTableBody.tsx
```diff
@@ -0,0 +1,120 @@
...
+                        >
+                            <Checkbox
+                                checked={isItemSelected}
+                                inputProps={{ "aria-labelledby": labelId }}
+                            />
+                        </TableCell>
+                        {columns.map(column => (
+                            <TableCell
+                                className={classes.cell}
+                                component="th"
```
SferaDev: I believe th is the default for Table ell, verify and remove if needed.

src/components/data-table/DataTableBody.tsx
```diff
@@ -0,0 +1,120 @@
...
+                        ))}
+                        <TableCell
+                            className={classes.cell}
+                            key={`${labelId}-actions`}
+                            padding="none"
+                            align={"center"}
+                        >
+                            <Tooltip title={i18n.t("Actions")} aria-label="actions">
+                                <IconButton onClick={event => contextualAction(row, event)}>
+                                    <MoreVertIcon />
```
SferaDev: Hide if no actions defined

src/components/data-table/DataTableHeader.tsx
```diff
@@ -0,0 +1,113 @@
...
+                                <span className={classes.visuallyHidden}>
+                                    {order === "desc" ? "sorted descending" : "sorted ascending"}
+                                </span>
+                            ) : null}
+                        </TableSortLabel>
+                    </TableCell>
+                ))}
+                <TableCell className={classes.cell} padding="none" align={"center"}>
+                    <IconButton>
+                        <ViewColumnIcon />
```
SferaDev: Implement hide/show columns

src/components/data-table/utils/sorting.tsx
```diff
@@ -0,0 +1,43 @@
...
+    if (b[orderBy] < a[orderBy]) {
+        return -1;
+    }
+    if (b[orderBy] > a[orderBy]) {
+        return 1;
+    }
+    return 0;
+}
+
+function sort(array: TableObject[], cmp: (a: TableObject, b: TableObject) => number) {
```
SferaDev: I do not particularly like this sorting technique, I used the default sorting of the original material-ui example but we can improve it.

src/components/data-table/utils/sorting.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+
+function getSorting<K extends keyof any>(
+    order: Order,
+    orderBy: K
+): (a: { [key in K]: number | string }, b: { [key in K]: number | string }) => number {
+    return order === "desc" ? (a, b) => desc(a, b, orderBy) : (a, b) => -desc(a, b, orderBy);
+}
+
+export function cleanObjects(
```
SferaDev: This should be renamed

src/components/data-table/ObjectsTable.tsx
```diff
@@ -0,0 +1,67 @@
...
+    detailFields?: ObjectsTableDetailField[];
+    initialSearch?: string;
+    onChangeSearch?(search: string): void;
+    onButtonClick?: Function;
+    buttonLabel?: ReactNode;
+    customFilters?: ReactNode;
+    hideSearchBox?: boolean;
+    forceSelectionColumn?: boolean;
+}
+
```
SferaDev: Add FAB and SearchBox

SferaDev:
This is mostly ready for review, it's a massive refactor and we might discuss changes in the implementation as we see them fit.

src/components/data-table/ContextualMenu.tsx
```diff
@@ -0,0 +1,74 @@
...
+                vertical: "center",
+                horizontal: "center",
+            }}
+            onClose={onClose}
+        >
+            {actions.map(action => (
+                <MenuItem
+                    className={classes.item}
+                    key={action.name}
+                    onClick={() => {
```
tokland: A bit too much for an inline function prop?

src/components/data-table/DataTable.tsx
```diff
@@ -0,0 +1,216 @@
...
+    const rowObjects = sortObjects(
+        rows,
+        uncontrolledPagination ? uncontrolledPagination : pagination,
+        uncontrolledSorting ? uncontrolledSorting : sorting
+    );
+
+    const primaryAction = _(availableActions).find({ primary: true }) || availableActions[0];
+    const allSelected = _.difference(rowObjects.map(row => row.id), selection).length === 0;
+    const enableMultipleAction = _.isUndefined(forceSelectionColumn)
+        ? !!_(availableActions).find({ multiple: true })
```
tokland: `_.some`

src/components/data-table/DataTableNotifications.tsx
```diff
@@ -0,0 +1,58 @@
...
+    },
+});
+
+export interface DataTableNotificationsProps {
+    messages: TableNotification[];
+    updateSelection?(selection: string[]): void;
+}
+
+export function DataTableNotifications(props: DataTableNotificationsProps) {
+    const { messages, updateSelection = () => {} } = props;
```
tokland: I just found out we have: `_.noop`

src/components/data-table/DataTablePagination.tsx
```diff
@@ -0,0 +1,47 @@
...
+    };
+
+    return (
+        <Pagination
+            rowsPerPageOptions={pagination.pageSizeOptions}
+            component="div"
+            count={pagination.total}
+            rowsPerPage={pagination.pageSize}
+            page={pagination.page - 1}
+            backIconButtonProps={{
```
tokland: Maybe to constant outside the function, we will also gain some vertical space.

src/components/data-table/utils/sorting.tsx
tokland: Yes, this a/b sorting is a bit clunky, better `_.sortBy`, or for a more sql aproach, `_.orderBy`.

src/components/data-table/DataTable.tsx
```diff
@@ -0,0 +1,216 @@
...
+            width: "100%",
+            minWidth: 750,
+        },
+        tablePagination: {
+            flex: "1 1 auto",
+        },
+    })
+);
+
+export interface DataTableProps {
```
tokland: Looks ok, but we will discuss it, the most important part of the PR is how we interface the public components.

src/components/data-table/DataTableNotifications.tsx
SferaDev: Nice!

tokland:
Minor: Go to 2nd page, search "qweqwe":

> Warning: Failed prop type: Material-UI: the page prop of a TablePagination is out of range (0 to 0, but page is 1).

src/components/data-table/types.d.ts
```diff
@@ -0,0 +1,47 @@
+import { ReactNode } from "react";
```
tokland: .d.ts extension in used as declaration files of JS code, we can drop the ".d"

src/components/data-table/DataTablePagination.tsx
SferaDev: Yeah, not sure if we want to keep the a11y props or not... We do not have them everywhere, I just mirrored the ones that material had in their examples.

src/components/data-table/types.d.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export interface TableObject {
+    id: string;
+    [key: string]: any;
+}
+
+export interface TableColumn {
+    name: string;
+    text: string;
+    sortable: boolean;
```
tokland: `sortable? boolean;` (default true?). And implement it, I think now they are always sortable.

tokland:
Not saying this is the moment to do it, but it would be nice to have the rows types. For that, an sketch:

```
export interface ObjectsTableProps<TableObject> extends DataTableProps<TableObject> {
 ...
}

export interface DataTableProps<TableObject> {
  rows: TableObject[];

export default function DataTable<TableObject>(props: DataTableProps<TableObject>) {
```

src/pages/TestPage.tsx
```diff
@@ -0,0 +1,107 @@
...
+//@ts-ignore
+import { Typography, Checkbox } from "@material-ui/core";
+
+function createData(
+    name: string,
+    calories: number,
+    fat: number,
+    carbs: number,
+    protein: number
+): any {
```
tokland: Let TS infer the return type. Also for `headCells`. This is how I noticed that property `sortable` should be optional :)

SferaDev:
> Not saying this is the moment to do it, but it would be nice to have the rows types. For that, an sketch:
> 
> ```
> export interface ObjectsTableProps<TableObject> extends DataTableProps<TableObject> {
>  ...
> }
> 
> export interface DataTableProps<TableObject> {
>   rows: TableObject[];
> 
> export default function DataTable<TableObject>(props: DataTableProps<TableObject>) {
> ```

Yeah, I originally began doing it a generic component but thought it would be too much for simple table creation.

Anyway I would like to do the following:

```
Either in d2-ui-components or d2-api we could create both a DataTable and ObjectsTable wrapper that manages the uncontrolled pagination and updates the sorting with server side object loading.
```

In that wrapper for d2-api objects we could properly type the rows too and provide the default TableObject for "naive" tables.

Right now the TableObject is anything that has a property called id (to make sure selection works fine).

adrianq:
This PR has been finally implemented in d2-ui-components here https://github.com/EyeSeeTea/d2-ui-components/pull/82

PR #202, opened 2019-11-06
Fix bug with filtered objects not properly loading paginated

### LINKED ISSUE #201: Sync rules do not paginate after first page

SferaDev:
### :pushpin: References

* **Issue:** Closes #201 

### :memo: Implementation

- Query all, filter and then create pager
- This needs to be ported to ``development-malaria``, does not happen in ``development`` because it was introduced with the sharing settings filtering feature.

### :bookmark_tabs: Others

- This change does not follow DRY as we have similar code also in the dataStore main class. However since all that code will be migrated to ``d2-api`` soon, the way we paginate over results and do the filters will completely change. At least we have it now working so WHO is able to use the app while we refactor this code with our new vision.

### DIFF
--- a/src/models/syncRule.ts
+++ b/src/models/syncRule.ts
@@ -158,8 +158,8 @@ export default class SyncRule {
         const globalAdmin = await isGlobalAdmin(d2);
         const userInfo = await getUserInfo(d2);
 
-        const data = await getPaginatedData(d2, dataStoreKey, filters, pagination);
-        const objects = _(data.objects)
+        const data = await getPaginatedData(d2, dataStoreKey, filters, { paging: false });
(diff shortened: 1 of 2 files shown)


adrianq:
It fixes the problem. While testing the fix I played around with the search filter and in particular the filter logic in getPaginatedData. Not important for now but we may want to re-visit this bit of code eventually:
```
.filter(k => typeof o[k] === "string")
.some(k => o[k].toLowerCase().includes(search.toLowerCase()))
```

adrianq:
@SferaDev something else I noticed https://github.com/EyeSeeTea/metadata-synchronization/issues/205 Let's talk tomorrow!

PR #215, opened 2019-11-13
[Bugfix] Move filtering into request and pre-calculate pager

### LINKED ISSUE #205: Last executed filter seem not to be working in development in the sync rules

SferaDev:
### :pushpin: References

* **Issue:** Closes #205 

### :memo: Implementation

- Re-calculate pagination after applying custom filters by tapping into the ``getPaginatedData`` call with a filtering function.

- This change does not apply DRY, but are temporal fixes until we are ready to move to the new approach using the new ``DataTable`` and ``d2Api.dataStore``. We will completely re-do this filtering calls and pagination calculation.

### DIFF
--- a/src/models/dataStore.ts
+++ b/src/models/dataStore.ts
@@ -72,7 +72,8 @@ export async function getPaginatedData(
     d2: D2,
     dataStoreKey: string,
     filters: TableFilters,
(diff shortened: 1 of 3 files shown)


adrianq:
This is already solved in a different branch

PR #222, opened 2019-11-19
[Feature] Data sync page

### LINKED ISSUE #168: Selection of DE's
Selection by:
- None (DE) by default
- DEG
- DEGS
- Dataset
- Program

Now:
- [x] Migrate to new table
- [x] Add column UID
- [x] Remove groups filter
- [x] Change filter and logic to show the aforementioned pieces of metadata

Later:
- [x] Allow collapse/expand
- [x] Both parent and child should be selectable
- [x] Pagination will only work with the parent items
- [x] Synchronization button

SferaDev:
### :pushpin: References

* **Issue:** Closes #168 (Part 2)

### :memo: Implementation

- Add wizard dialog and sync dialog into "Data Sync" page
- Move selection into "on-demand" sync rule following @xurxodev's code
- Query the API for children rows

### :art: Screenshots

image
image

- This just creates the UI that will be used for "data sync".
- We have created a subset of issues that have less priority for now: #223 #225 #226 #227

### :bookmark_tabs: Others

https://github.com/EyeSeeTea/d2-ui-components/pull/94


SferaDev:
@tokland When you review this, I have not yet created a model per-se for data synchronization because we have not began with the logic here. I plan to create a class that handles the synchronization and the types/fields that are now in the ``DataPage`` component. This new model class will be much needed to allow the transforms required to make DataSets and Programs work properly.

PR #278, opened 2019-12-26
[Feature] Change grid for maintenance layout

### LINKED ISSUE #257: Landing Page Improvements
- [x] Change 'synchronization' to 'sync'
- [x] Make 'metadata', 'aggregated' and 'events' bold or something along these lines
- [ ] Review icons. Idea from the client: use clock icon for rules and something else different for the history sections
- [ ] Use another icon for the instances. A server icon?
- [x] Change  'Instance Configuration' to 'Destination Instance Settings'
- [x] Make even more clear the 3 blocks: Metadata, Aggregated and Events
- [x] Change 'Aggregated Synchronization' to 'Aggregated Data Sync'

adrianq:
### :pushpin: References

* **Issue:** Closes #257 

### :memo: Implementation

-

### :bookmark_tabs: Others

[Landing Page Improvements](https://app.gitkraken.com/glo/board/XaBKjXlqfAAPh4Tg/card/XfNK6I40NwAQ8ZuX)


adrianq:
Thanks @SferaDev for the detailed review! I have followed your suggestion and move the instance settings to the bottom. I agree it looks much better. I have added some extra left margin. It is not gonna solve how this looks in big screens but I believe it looks a bit nicer.

image

Something to discuss, in the official maintenance app if click on the '+' to create a metadata object and then press the back button you end up in the list section of the object. Right now, we, however, send the user back to the landing page. I think it makes sense to mimic the maintenance behaviour, right?

Btw, something I don't particularly like about the latest changes: https://github.com/EyeSeeTea/metadata-synchronization/pull/278/files#diff-817a7b231e0f2d3825855c9c49f47b2fR58 I am passing the dataTest variable to the MenuCard component. I guess there is a much more elegant solution.

I have also fixed the travis tests. There was a bug in travis when reaching a maximum number of files to watch. It seems to be fixed with this https://github.com/EyeSeeTea/metadata-synchronization/pull/278/commits/4f322ec6685b8a4bb44d834f83b8d426cac5b2e9 I had to released a beta version of the d2-ui-comp to make tests pass on travis.

SferaDev:
> Something to discuss, in the official maintenance app if click on the '+' to create a metadata object and then press the back button you end up in the list section of the object. Right now, we, however, send the user back to the landing page. I think it makes sense to mimic the maintenance behaviour, right?

Makes sense to me too. Right now though we have to routes string hardcoded and I would like to have them referenced from the Root file.

> Btw, something I don't particularly like about the latest changes: https://github.com/EyeSeeTea/metadata-synchronization/pull/278/files#diff-817a7b231e0f2d3825855c9c49f47b2fR58 I am passing the dataTest variable to the MenuCard component. I guess there is a much more elegant solution.

I didn't check the test code yet, but normally it is not needed to send the data-test to the children component. Just use it in the parent and it should be good to go. So instead of sending the prop and reading down the tree, you can set it in LandingPage and ignore it in MenuCard.

> I have also fixed the travis tests. There was a bug in travis when reaching a maximum number of files to watch. It seems to be fixed with this [4f322ec](https://github.com/EyeSeeTea/metadata-synchronization/commit/4f322ec6685b8a4bb44d834f83b8d426cac5b2e9) I had to released a beta version of the d2-ui-comp to make tests pass on travis.

The commands on the pre-install are Linux (and maybe even bash) dependant and will likely fail on other environments. I agree that the patch of increasing the max inodes is useful specially when working with VSCode but I wouldn't add it there to maintain compatibility. Just adding it to the travis.yml would fix the CI and still work on other platforms and environments.

Issue #289, opened 2019-12-31
User feedback Alpha release


SferaDev:
High Priority

Low Priority
- [ ] Sync Rules with multiple instances but different sharing settings [(Issue)](https://github.com/EyeSeeTea/metadata-synchronization-blessed/issues/20)  -> Created issue #300
- [ ] Add sharing settings in wizard for sync rule creation [(Issue)](https://github.com/EyeSeeTea/metadata-synchronization-blessed/issues/22) -> Created issue #300

Duplicated
- [x] Cannot unselect children from parent [(Issue)](https://github.com/EyeSeeTea/metadata-synchronization-blessed/issues/27) -> Closing this as it is already captured in our repo and reported as a missing feature. See https://github.com/EyeSeeTea/metadata-synchronization/issues/238

Under testing
- [x] Filters overlapping [(Issue)](https://github.com/EyeSeeTea/metadata-synchronization-blessed/issues/25)
- [x] Pagination cannot be clicked [(Issue)](https://github.com/EyeSeeTea/metadata-synchronization-blessed/issues/26)

Done
- [x] Wrong item selected [(Issue)](https://github.com/EyeSeeTea/metadata-synchronization-blessed/issues/23) [(PR)](https://github.com/EyeSeeTea/d2-ui-components/pull/103) 
- [x] Scheduling custom option does not stay selected [(Issue)](https://github.com/EyeSeeTea/metadata-synchronization-blessed/issues/24) [(PR)](https://github.com/EyeSeeTea/metadata-synchronization/pull/269) 
- [x] Landing page description? [(Issue)](https://github.com/EyeSeeTea/metadata-synchronization-blessed/issues/31) [(PR)](https://github.com/EyeSeeTea/metadata-synchronization/pull/295)
- [x] Hide plus button in landing page if user hasn't permissions [(PR)](https://github.com/EyeSeeTea/metadata-synchronization/pull/295)
- [x] Replicate sync rule [(Issue)](https://github.com/EyeSeeTea/metadata-synchronization-blessed/issues/21) [(PR)](https://github.com/EyeSeeTea/metadata-synchronization/pull/290)
- [x] Instances list page, default action: Edit [(Issue)](https://github.com/EyeSeeTea/metadata-synchronization-blessed/issues/32) [(PR)](https://github.com/EyeSeeTea/metadata-synchronization/pull/290)
[...]

PR #290, opened 2019-12-31
Add replicate and update default table actions

SferaDev:
### :pushpin: References

* **Issue:** Works in #289 

### :memo: Implementation

- Add replicate contextual action to "SyncRules" and "Instances"
- Update default action in "Instances" page
- Update default action in "History" page

### :art: Screenshots

image
image
image
image


adrianq:
@SferaDev This PR looks good. I have just discovered another bug (https://github.com/EyeSeeTea/metadata-synchronization/issues/297) but it doesn't look like it has to do with the replicate function. It is probably a bug we introduce together with the cat combo step. Can you please have a look at it today? I have the impression it may be simple to solve.

PR #295, opened 2020-01-01
Fix/landing page

adrianq:
### :pushpin: References

* **Issue:** Closes part of #289?

### :memo: Implementation
- Add proper description to landing page objects
- Show creation action link ('+' button) in the landing page only if the user has the configuration role

### :art: Screenshots
image

### :bookmark_tabs: Others


src/pages/landing/MenuCard.tsx
```diff
@@ -13,7 +13,7 @@ import React from "react";
 export interface MenuCardProps {
     name: string;
     description?: string;
-    addAction?: () => void;
+    addAction?: (() => void) | null;
```
SferaDev: For simplicity, instead of "null" we can return "undefined" when the conditions are false and keep the definition as it was before.

adrianq: Yep, it was my first approach as well but then I thought it was maybe cleaner to return a more explicit null. Anyway, will change it and merge. Thanks!

PR #302, opened 2020-01-03
use docker for testing

adrianq:
### :pushpin: References

* **Issue:** Closes #?

### :memo: Implementation

-

### :bookmark_tabs: Others

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,14 +28,12 @@ before_install:
 install:
   - yarn install --frozen-lockfile
   - yarn cy:verify
-  # - yarn build
-  - yarn build-webapp
+  - yarn build
 script:
-  # - PORT=8081 REACT_APP_DHIS2_BASE_URL=http://localhost:8080 REACT_APP_CYPRESS=true yarn start &
+  - PORT=8081 REACT_APP_DHIS2_BASE_URL=http://localhost:8080 REACT_APP_CYPRESS=true yarn start &
   - yarn wait-on http-get://localhost:8080
-  - curl -F file=@metadata-synchronization.zip -X POST -u $CYPRESS_DHIS2_AUTH "http://localhost:8080/api/apps" 
-  # - yarn wait-on http-get://localhost:8081
-  - CYPRESS_EXTERNAL_API=http://localhost:8080 CYPRESS_ROOT_URL=http://localhost:8080 yarn cy:e2e:run
+  - yarn wait-on http-get://localhost:8081
+  - CYPRESS_EXTERNAL_API=http://localhost:8080 CYPRESS_ROOT_URL=http://localhost:8081 yarn cy:e2e:run
   - kill $(jobs -p) || true
 addons:
   apt:
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ install:
 script:
   # - PORT=8081 REACT_APP_DHIS2_BASE_URL=http://localhost:8080 REACT_APP_CYPRESS=true yarn start &
   - yarn wait-on http-get://localhost:8080
-  - curl -F file=@metadata-synchronization.zip -X POST -u '$CYPRESS_DHIS2_AUTH' "http://localhost:8080/api/apps" 
+  - curl -F file=@metadata-synchronization.zip -X POST -u "$CYPRESS_DHIS2_AUTH" "http://localhost:8080/api/apps" 
   # - yarn wait-on http-get://localhost:8081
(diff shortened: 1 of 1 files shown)


SferaDev:
@adrianq @tokland I have been thinking about the "production" and "development test execution.

While it Is true that testing it in production allows us to verify the bundle is properly built, running them in development allows us to create a more robust testing environment. 

Let me explain my idea for how we could improve our E2E tests. Since it adds overhead I would strongly suggest to keep our environment in development for testing. 

**Spend the minimum time writing test**

There are tools for recording user actions that generates a Cypress compatible test script. These tools have improved a lot since we originally checked them as cypress has gained more popularity. However you still need to provide data-test ids if you want to created meaningful and readable outputs. 

**Add Istanbul for code coverage**

Whilst code coverage is not a good indicator of how good is a test suite, it allows to verify what parts of the code base need more tests. I think we could add it to the platform. Note that this kind of instrumentation only works in development mode. 

**Writing data-test-id**

I would like to build a function that could be reused in our rest of applications that traverses the react tree to create data-test information. This would be a huge overhead that it's acceptable in development but that would completely be ignored in production.

My current idea is to recursively traverse the tree and for each node follow the following rules:

1. If data-test-id is already provided, do mothing
2. If react.key is provided use it to create data-test-id
3. If none of the above use, parent data-test-id + name of the component + numeric identifier ``sync-rule-list-div-4``.

I believe this way we can speed up a lot the process and only define keys where needed (react already does complain about it) and use the auto-generation handle the rest.

adrianq:
@SferaDev I have reverted the changes and now we are using the development bundle. We still need to discuss the best approach for testing but this is already better than what we have before. I would like to merge the code as-is. We can always iterate and improve it at a later stage. I am adding you as a reviewer. Please have a quick look at it and let me know if I can go ahead and merge it... and btw, thanks for the thoughtful comment!

tokland:
Cypress macros: The package I checked at the ending of vaccination-app (cypress-recorder) seemed pretty limited, which ones have you found? I am all for writing easier macro-like integration tests, even though at the moment I was a bit worried about the limitations on the assertions we could do.

Istanbul: Sure, although not having a goal of 100% test coverage, it may be kind of disappointing to see the metrics :)

> Writing data-test-id

Cool idea! I am just a bit worried about one thing: does it slow significantly the rendering? losing confidence in the render speed in DEV (not knowing if it's the app or this post-processing) may be a problem.

> Since it adds overhead I would strongly suggest to keep our environment in development for testing.

Not sure I get you, but what you propose should be orthogonal to using docker/deployed-app-zip, right? We could always create a PRO build with data-test-id attributes (setting an ENV variable or whatever)

SferaDev:
> Cypress macros: The package I checked at the ending of vaccination-app (cypress-recorder) seemed pretty limited, which ones have you found? I am all for writing easier macro-like integration tests, even though at the moment I was a bit worried about the limitations on the assertions we could do.

Yeah there're a couple of "good" options.

> Istanbul: Sure, although not having a goal of 100% test coverage, it may be kind of disappointing to see the metrics :)

Agreed. 

> > Writing data-test-id
> 
> Cool idea! I am just a bit worried about one thing: does it slow significantly the rendering? losing confidence in the render speed in DEV (not knowing if it's the app or this post-processing) may be a problem.

It does not slow down too much, at least on my PC it wasn't noticeable. But it surely slows down so we should keep it out PROD. 

> > Since it adds overhead I would strongly suggest to keep our environment in development for testing.
> 
> Not sure I get you, but what you propose should be orthogonal to using docker/deployed-app-zip, right? We could always create a PRO build with data-test-id attributes (setting an ENV variable or whatever)

Right, that's true.

SferaDev:
The tests seem to work fine. We can merge this as-is and discuss the "development" and "production" later on.

adrianq:
@SferaDev What libs do you have in mind for cypress recording? I would like to have a look at them before our testing meeting so we can make the most of it. Thanks!

SferaDev:
> @SferaDev What libs do you have in mind for cypress recording? I would like to have a look at them before our testing meeting so we can make the most of it. Thanks!

I have two in mind. I couldn't make them retrieve the ``data-test`` ids yet (even though at least one of them should support it). But both of them are OSS and with a minor tweak we could make it work nicely.

**FD Cypress Recorder**

https://chrome.google.com/webstore/detail/fd-cypress-recorder/amleackadkomdccpbfginhnecfhhognj

image

**Cypress Scenario Recorder**

https://chrome.google.com/webstore/detail/cypress-scenario-recorder/fmpgoobcionmfneadjapdabmjfkmfekb

image

PR #313, opened 2020-01-08
[Feature] Misc changes for user feedback

### LINKED ISSUE #305: Currently everyday means mid-day, make it midnight by default

### LINKED ISSUE #306: The summary of a sync rule is not working when editing a sync rule
Blessed issue =>  https://github.com/EyeSeeTea/metadata-synchronization-blessed/issues/33

### LINKED ISSUE #308: When exporting the json file of a sync rule, prepend the name with the sync rule name

SferaDev:
### :pushpin: References

* **Issue:** Closes #305
* **Issue:** Closes #306
* **Issue:** Closes #308 

### :memo: Implementation

- Move JSON download to a new utils file
- Prepend sync rule name using lowercase kebab-case ``foo-bar-sync-date``.
- Rename internal "on-demand" name to "manual" so when downloading a file it looks like ``manual-aggregated-sync-date``.
- Change all cron expressions to be ran during midnight by default.
- Add moment wrapper for data sync rule start and end dates

### :art: Screenshots

image
image

### :bookmark_tabs: Others

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "@babel/core": "^7.7.5",
     "@babel/preset-typescript": "^7.7.4",
     "@types/cryptr": "^4.0.0",
+    "@types/file-saver": "^2.0.1",
     "@types/jest": "^24.0.23",
     "@types/lodash": "^4.14.149",
     "@types/node": "^12.12.14",
--- a/src/components/sync-wizard/common/SchedulerStep.jsx
+++ b/src/components/sync-wizard/common/SchedulerStep.jsx
@@ -8,11 +8,11 @@ import { Toggle } from "../../toggle/Toggle";
 import isValidCronExpression from "../../../utils/validCronExpression";
 
 const cronExpressions = [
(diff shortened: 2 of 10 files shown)


package.json
```diff
@@ -74,6 +74,7 @@
     "@babel/core": "^7.7.5",
     "@babel/preset-typescript": "^7.7.4",
     "@types/cryptr": "^4.0.0",
+    "@types/file-saver": "^2.0.1",
```
adrianq: @SferaDev  Why is file saver now a dev dependency? Did you do this on purpose?

adrianq: @SferaDev I am merging this because everything seems to be working fine and I want to finish reviewing the pending PRs but please let me know about the file saver dependency. We can always fix it in another PR if needed. Thanks!

SferaDev: This is the type definitions. Previously we only used the file saver lib on JS files but with this refactor we moved the code to a TS one. The type definitions are now needed and we include them always as dev dependencies.

PR #323, opened 2020-01-24
Feature/include exclude selection

### LINKED ISSUE #216: Add option to include/exclude dependencies on sync

xurxodev:
* **Issue:** Closes #216 

### :memo: Implementation

- I have created a toggle component to use in the new step
- I have created the IncludeExcludeSelectionStep component
- I have created two new fields in SyncRule: useDefaultIncludeExclude and metadataIncludeExcludeRules
- The IncludeExcludeSelectionStep the first time has the use default configuration toggle to true. With this configuration, the synchronization to use exclude include rules hardcoded in d2models. (the fields in sync rule for this configuration are useDefaultIncludeExclude= true and metadataIncludeExcludeRules= undefined)
- When the use default configuration toggle is set to false by the user then SyncRule assigns MetadataIncludeExcludeRules like exclude include rules hardcoded in the model for every metadata type according to selected metadata in the previous step. Then the user can realized modifications in include and exclude rules for every metadata type.
- When metadataId changes IncludeExcludeSelectionStep is initialized to default configuration because is possible that rules added to exclude, include for a type has no sense because this type now does not exist in metadataIds.
- I have created friendly names to exclude include rules in a dictionary and has been created in the IncludeExcludeSelectionStep component because it’s a presentation layer concern.
- I have added IncludeExcludeSelectionStep info in the save step
- I have added logic related to parent rules. When a rule is moved from excluded to included the parents if exists, must to be moved also: for example if the user moves from exclude to include organisationUnitGroups.organisationUnitGroupSets.attributes then organisationUnitGroups.organisationUnitGroupSets and organisationUnitGroups must be moved to include because is not possible sync attributes of organisationUnitGroupSets if this organisationUnitGroupSets does not exist in target instance for example.
- I have added logic related to children rules. When a rule is moved from include to excluded the children if exists, must be moved also: for example if the user moves from include to exclude organisationUnitGroupSets then organisationUnitGroups.organisationUnitGroupSets and organisationUnitGroups.organisationUnitGroupSets.attributes must be moved to avoid the explained problems in the previous point.
 
To found parents and children I have not created extra arrays or dictionaries, I have used the current exclude include rules in models. 

For example:
Rule: organisationUnitGroups.organisationUnitGroupSets.attributes
Parents: [organisationUnitGroups.organisationUnitGroupSets, organisationUnitGroups]

I have created the extractParentsFromRule function to this requirement and tests to validate it.

Rule: organisationUnitGroups
Children: either rule in its model that contains “organisationUnitGroups.”

### :art: Screenshots

include-exclude-step
save-step

**Use case 1**:  create a new sync rule and use default configuration toggle in the include exclude selection step should have marked by default 
**Use case 2**:  When the toggle is marked to false, then for every type of metadataIds should be assigned the default exclude include rules hardcoded in the models and the user should can realize changes.
**Use case 3**:  When metadataIds are selected and exclude include rules also, if we go back to metadata step and realize changes, then use default configuration toggle in exclude include selection step should be initialized to true.
**Use case 4**:  in exclude include selection step when a rule is moved from exclude to include, parents of the rule, if exists, should be moved automatically also.
**Use case 5**:  in exclude include selection step when a rule is moved from exclude to include and its parent exists previously in the include list should not create duplicates in the include list.
**Use case 6**:  in exclude include selection step when a rule and parents are selected and moved from exclude to include, should not exists duplicates in the include list.
**Use case 7**:  in exclude include selection step when a rule is moved from include to exclude, children of the rule, if exists, should be moved automatically also.
**Use case 8**:  in exclude include selection step when a rule and children are selected and moved from include to exclude, should not exists duplicates in the exclude list.
**Use case 9**:  when a sync rule is executed with use default configuration toggle in exclude include selection step mark to true then should to synchronize only relationships in include list for every model.
**Use case 10**:  when a sync rule is executed with use default configuration toggle in exclude include selection step mark to false then should to synchronize using hardcoded exclude include list in every model class.
**Use case 11**:  when a sync rule contains use default configuration toggle in the exclude include selection step mark to true then save step should contain only toggle information.
**Use case 12**:  when a sync rule contains use default configuration toggle in the exclude include selection step mark to false then save step should contain the toggle information and exclude include rules for every selected model type in metadata step.
**Use case 13**:  OnDemand synchronizations should work the same way before

### :bookmark_tabs: Others

This PR contains a lot of logic, this logic has been created in the model SyncRule to have isolated the logic, this do possible unit tests creation.

To create the logic of this issue have used TDD, as a consequence, many new tests have been added while the code was created. The use of TDD is slower creating the solution the first time but I hope on the future these tests simplify the future refactors and modifications.

Many use cases to test to verify this PR are testing using unit testing, if you found some case that is not implemented as a unit test, please let me know.

**Note**: SyncRule model is a class with many responsibilities and getting bigger and bigger
, in the future when we have time, a refactor to split the responsibilities may be a good option, but I think that I need to know better your development architecture to propose improvements.


src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
```diff
@@ -0,0 +1,236 @@
...
+    };
+
+    const getRuleOptions = () => {
+        const allRules = getAllRules();
+
+        return allRules.map(rule => ({
+            value: rule,
+            text: includeExcludeRulesFriendlyNames[rule]
+                ? includeExcludeRulesFriendlyNames[rule]
+                : rule,
```
SferaDev: [minor] ``includeExcludeRulesFriendlyNames[rule] ?? rule``

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
```diff
@@ -0,0 +1,236 @@
...
+interface ModelSelectItem {
+    name: string;
+    id: string;
+}
+
+interface friendlyRulesNamesDiccionary {
+    [metadataType: string]: string;
+}
+
+const includeExcludeRulesFriendlyNames: friendlyRulesNamesDiccionary = {
```
SferaDev: Not sure if it's the time of doing this abstraction but could we build the dictionary with a ``_(includeExcludeRule).split(".").reverse().map(getModelDisplayName).join(" of ").value()``

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
```diff
@@ -0,0 +1,236 @@
...
+    "validationRuleGroups.attributes": "Attributes of validation rule groups",
+};
+
+const IncludeExcludeSelectionStep: React.FC<IncludeExcludeSelectionStepProps> = ({
+    d2,
+    syncRule,
+    onChange,
+}) => {
+    const [modelSelectItems, setModelSelectItems] = useState<Array<ModelSelectItem>>([]);
+    const [models, setModels] = useState<Array<typeof D2Model>>([]);
```
SferaDev: [minor] Type definitions could be ``ModelSelectItem[]`` and ``(typeof D2Model)[]``. I believe tokland and I are using this in favor of ``Array<>``.

src/models/syncRule.ts
```diff
@@ -133,11 +149,146 @@ export default class SyncRule {
...
+
+        return SyncRule.build({
+            ...this.syncRule,
+            builder: {
+                ...this.syncRule.builder,
+                useDefaultIncludeExclude: false,
+                metadataIncludeExcludeRules,
+            },
+        });
+    }
```
SferaDev: These two methods could be joined and/or simplified? We have two methods to touch the same two properties.

I'm also not sure if I would make the toggle independent of the customized rules, so the user does not loose them when enabling or disabling the behavior.

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
```diff
@@ -0,0 +1,236 @@
...
+                : syncRule.markToNotUseDefaultIncludeExclude(models)
+        );
+    };
+
+    const changeModelName = (event: any) => {
+        setSelectedType(event.target.value);
+    };
+
+    const changeInclude = (includeRules: any) => {
+        const type: string = selectedType ?? "";
```
SferaDev: [very minor] type is inferred as string from ``selectedType ?? ""``

src/logic/synchronization.ts
```diff
@@ -87,11 +92,17 @@ export async function* startSynchronization(
...
                 type,
                 ids: metadata[type].map(e => e.id),
-                excludeRules: myClass.getExcludeRules(),
-                includeRules: myClass.getIncludeRules(),
+                excludeRules: metadataIncludeExcludeRules
+                    ? metadataIncludeExcludeRules[metadataType].excludeRules.map(_.toPath)
+                    : myClass.getExcludeRules(),
+                includeRules: metadataIncludeExcludeRules
+                    ? metadataIncludeExcludeRules[metadataType].includeRules.map(_.toPath)
+                    : myClass.getIncludeRules(),
```
SferaDev: I would use ``useDefaultIncludeExclude`` instead of undefined-checking here.

src/models/syncRule.ts
```diff
@@ -42,6 +48,14 @@ export default class SyncRule {
         return this.syncRule.builder.metadataIds;
     }
 
+    public get useDefaultIncludeExclude(): boolean {
+        return this.syncRule.builder.useDefaultIncludeExclude;
+    }
+
+    public get metadataExcludeIncludeRules(): MetadataIncludeExcludeRules | undefined {
```
SferaDev: Since we have ``useDefaultIncludeExclude``, it would make sense to disallow it being undefined.

Note: Existing rules will lead to an error when not finding it. We usually add ``?? {}`` for this particular case.

src/models/syncRule.ts
```diff
@@ -42,6 +48,14 @@ export default class SyncRule {
         return this.syncRule.builder.metadataIds;
     }
 
+    public get useDefaultIncludeExclude(): boolean {
+        return this.syncRule.builder.useDefaultIncludeExclude;
```
SferaDev: Note: Existing rules will lead to an error when not finding it. We usually add ``?? true`` for this particular case.

src/models/syncRule.ts
```diff
@@ -133,11 +149,146 @@ export default class SyncRule {
...
+            return SyncRule.build({
+                ...this.syncRule,
+                builder: {
+                    ...this.syncRule.builder,
+                    metadataIds,
+                    useDefaultIncludeExclude: true,
+                    metadataIncludeExcludeRules: undefined,
+                },
+            });
+        }
```
SferaDev: Could we assume that if ``updateMetadataIds`` has been called, the metadataIds did change?

If we are receiving false-positives that's a bug and we shouldn't tackle with an ``isEqual`` on model.

I would particularly do the following without any other check:

```ts
            return SyncRule.build({
                ...this.syncRule,
                builder: {
                    ...this.syncRule.builder,
                    metadataIds,
                    useDefaultIncludeExclude: true,
                    metadataIncludeExcludeRules: {},
                },
            });
```

src/types/synchronization.d.ts
```diff
@@ -2,11 +2,22 @@ import { MetadataImportResponse, MetadataImportStats } from "./d2";
 import SyncReport from "../models/syncReport";
 
 export interface SynchronizationBuilder {
+    metadataIncludeExcludeRules: MetadataIncludeExcludeRules | undefined;
```
SferaDev: Same as before ``MetadataIncludeExcludeRules`` instead of ``MetadataIncludeExcludeRules | undefined``

.eslintrc.js
```diff
@@ -17,6 +17,7 @@ module.exports = {
         "@typescript-eslint/indent": "off",
         "@typescript-eslint/member-delimiter-style": "off",
         "@typescript-eslint/type-annotation-spacing": "off",
+        "@typescript-eslint/no-use-before-define": "off",
```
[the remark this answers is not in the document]
xurxodev: @SferaDev yes I have disabled this lint rule for testing files. It's unreadable and a nuisance having to scroll to read the tests in large files where there are helper functions. This change was suggested to me by @tokland. In the project-app, he disabled it also.

Maybe a point to talk in the Monday meeting.

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
xurxodev: I have doubts if it would be grammatically correct the dictionary result items using this code, anyway, friendly names for rules are pending @adrianq revision, I think after revision it might be time to think if it makes sense to try to obtain the dictionary using an algorithm.

src/models/syncRule.ts
xurxodev: Two methods are necessary because the behaviour and arguments are different in every case. 
Customization rules are dependent on toggle because if the toggle is marked to true, the user can not customize and should lose previous customizations. And when the toggle is marked to false, It's necessary to set metadataIncludeExcludeRules as hardcoded in models.
This behaviour allows to synchronization process using hardcoded exclude include rules in models when metadataIncludeExcludeRules are undefined (for example on-demand sync)

src/logic/synchronization.ts
xurxodev: The really important field to sync is metadataIncludeExcludeRules, the useDefaultIncludeExclude field is a helper field for when you are creating the sync rule.

Using useDefaultIncludeExclude field is not sufficient to decide using hardcoded rules in models or metadataIncludeExcludeRules because metadataIncludeExcludeRules can be undefined.

there two options:
- current 
       excludeRules: metadataIncludeExcludeRules
                    ? metadataIncludeExcludeRules[metadataType].excludeRules.map(_.toPath)
                    : myClass.getExcludeRules(),
                includeRules: metadataIncludeExcludeRules
                    ? metadataIncludeExcludeRules[metadataType].includeRules.map(_.toPath)
                    : myClass.getIncludeRules(),
- with useDefaultIncludeExclude 
           excludeRules: metadataIncludeExcludeRules && useDefaultIncludeExclude
                    ? metadataIncludeExcludeRules[metadataType].excludeRules.map(_.toPath)
                    : myClass.getExcludeRules(),
                includeRules: metadataIncludeExcludeRules && useDefaultIncludeExclude
                    ? metadataIncludeExcludeRules[metadataType].includeRules.map(_.toPath)
                    : myClass.getIncludeRules(),

I think the current option is more simpler.

src/models/syncRule.ts
xurxodev: When useDefaultIncludeExclude is equal to true the sync rule does not contains include exclude rules to avoid duplicate and innecessary rules stored in dhis2 store and to avoid bugs if default rules changes on the future.

During the synchronization, the rules will be hardcoded in models if MetadataIncludeExcludeRules is undefined

src/models/syncRule.ts
xurxodev: No, we can not assume when updateMetadataIds has been called, the metadataIds did change. Currently, this method is invoked when the user to navigate for steps without making any changes. So this logic must be controlled in the model for several reasons: protect its invariants (model cannot reach an inconsistent state), code reuse, testing etc..

src/types/synchronization.d.ts
xurxodev: Same as before also :)

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
xurxodev: ok I'll do it

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
xurxodev: ok I'll do it

src/models/syncRule.ts
xurxodev: ok I'll do it

.eslintrc.js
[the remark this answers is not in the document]
SferaDev:
```
"no-use-before-define": ["error", { "functions": false, "classes": true, "variables": true }]
```

src/components/rules-creation-page/SyncRulesWizard.jsx
```diff
@@ -36,6 +37,14 @@ class SyncRulesWizard extends React.Component {
             description: undefined,
             help: undefined,
         },
+        {
+            key: "include-exclude-selection",
+            label: i18n.t("Include Exclude Selection"),
+            component: IncludeExcludeStep,
+            validationKeys: ["includeExclude"],
```
SferaDev: [minor] Why don't we use metadataIncludeExclude for the validation key? Here we only have Metadata but when porting to data-Sync could be confusing.

src/models/syncRule.ts
```diff
@@ -203,6 +358,7 @@ export default class SyncRule {
                       }
                     : null,
             ]),
+            includeExclude: [],
```
SferaDev: [minor] Why don't we use metadataIncludeExclude for the validation key? Here we only have Metadata but when porting to data-Sync could be confusing.

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
```diff
@@ -0,0 +1,236 @@
...
+}
+
+interface ModelSelectItem {
+    name: string;
+    id: string;
+}
+
+interface friendlyRulesNamesDiccionary {
+    [metadataType: string]: string;
+}
```
SferaDev: Since there is no usage of ``friendlyRulesNamesDiccionary`` out of ``includeExcludeRulesFriendlyNames`` we don't need to create an interface. It could be simplified to:

```
const includeExcludeRulesFriendlyNames: {
    [metadataType: string]: string;
} = {
...
}
```

Also since it's not an overly complex type, maybe TS already correctly infers it.

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
```diff
@@ -0,0 +1,236 @@
...
+                onValueChange={changeUseDefaultIncludeExclude}
+            />
+            {!syncRule.useDefaultIncludeExclude && (
+                <div
+                    style={{
+                        display: "flex",
+                        flexDirection: "column",
+                        alignItems: "flex-start",
+                        marginTop: "20px",
+                    }}
```
SferaDev: All the styling should be done through material-ui, sometimes in development looks good with inlined styles but on production when the DOM classes are optimized there're problems.

You can check examples on other components:

```
const useStyles = makeStyles({
    className: {
        // Styling
    },
});
```

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
```diff
@@ -0,0 +1,236 @@
...
+};
+
+const IncludeExcludeSelectionStep: React.FC<IncludeExcludeSelectionStepProps> = ({
+    d2,
+    syncRule,
+    onChange,
+}) => {
+    const [modelSelectItems, setModelSelectItems] = useState<ModelSelectItem[]>([]);
+    const [models, setModels] = useState<typeof D2Model[]>([]);
+    const [selectedType, setSelectedType] = useState<string | undefined>();
```
SferaDev: Dropdown's ``onChange`` does not return ``undefined``, instead ``""`` is returned. We can simplify it as ``useState<string>("")``.

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
```diff
@@ -0,0 +1,236 @@
...
+                        flexDirection: "column",
+                        alignItems: "flex-start",
+                        marginTop: "20px",
+                    }}
+                >
+                    <Dropdown
+                        key={"model-selection"}
+                        items={modelSelectItems}
+                        onChange={changeModelName}
+                        value={selectedType ? selectedType : ""}
```
SferaDev: If we simplify ``selectedType`` to be only string, this could be: ``value={selectedType}``.

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
```diff
@@ -0,0 +1,236 @@
...
+        return selectedType && syncRule.metadataExcludeIncludeRules
+            ? syncRule.metadataExcludeIncludeRules[selectedType].includeRules
+            : [];
+    };
+
+    const getExcludeRules = () => {
+        return selectedType && syncRule.metadataExcludeIncludeRules
+            ? syncRule.metadataExcludeIncludeRules[selectedType].excludeRules
+            : [];
+    };
```
SferaDev: This could be simplified to:

```
const { includeRules = [], excludeRules = [] } = syncRule.metadataExcludeIncludeRules[selectedType] || {};
```

Also in functional components there's no need to wrap them in a function, this line can be placed at the beggining and re-used on ``changeInclude`` and the render return.

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
```diff
@@ -0,0 +1,236 @@
...
+        const allRules =
+            selectedType && syncRule.metadataExcludeIncludeRules
+                ? [
+                      ...syncRule.metadataExcludeIncludeRules[selectedType].excludeRules,
+                      ...syncRule.metadataExcludeIncludeRules[selectedType].includeRules,
+                  ]
+                : [];
+
+        return allRules;
+    };
```
SferaDev: Following the optimization of https://github.com/EyeSeeTea/metadata-synchronization/pull/323/files#r371647259

```
const allRules = [ ...includeRules, ...excludeRules ];
```

Also this does not need to be wrapped in a function, just defined after ``includeRules`` and ``excludeRules``.

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
```diff
@@ -0,0 +1,236 @@
...
+    const getRuleOptions = () => {
+        const allRules = getAllRules();
+
+        return allRules.map(rule => ({
+            value: rule,
+            text: includeExcludeRulesFriendlyNames[rule]
+                ? includeExcludeRulesFriendlyNames[rule]
+                : rule,
+        }));
+    };
```
SferaDev: Following the comments of:

- https://github.com/EyeSeeTea/metadata-synchronization/pull/323/files#r371647259
- https://github.com/EyeSeeTea/metadata-synchronization/pull/323/files#r371648385

This could be moved outside of a wrapper and:

```
const ruleOptions = allRules.map(rule => ({ value: rule, text: includeExcludeRulesFriendlyNames[rule] || rule }));
```

NOTE: Why moving them outside of wrappers. All the variables you reference inside of an arrow function in JS are "stored" with the value they had when the arrow function was defined. If they change over time (and does not have the same object reference) they won't get updated. Even though these were simple arrow functions and they might not have any problem, this approach potentially leads to unwanted bugs that are hard to debug.

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
```diff
@@ -0,0 +1,236 @@
...
+                    />
+
+                    {selectedType && (
+                        <div style={{ width: "100%" }}>
+                            <MultiSelector
+                                style={{ width: "100%" }}
+                                d2={d2}
+                                height={300}
+                                onChange={changeInclude}
+                                options={getRuleOptions()}
```
SferaDev: Following https://github.com/EyeSeeTea/metadata-synchronization/pull/323/files#r371651073

```
options={ruleOptions}
```

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
```diff
@@ -0,0 +1,236 @@
...
+                                selected={getIncludeRules()}
```
SferaDev: Following https://github.com/EyeSeeTea/metadata-synchronization/pull/323/files#r371647259

```
selected={includeRules}
```

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
```diff
@@ -0,0 +1,236 @@
...
+    const changeUseDefaultIncludeExclude = (useDefault: boolean) => {
+        onChange(
+            useDefault
+                ? syncRule.markToUseDefaultIncludeExclude()
+                : syncRule.markToNotUseDefaultIncludeExclude(models)
+        );
+    };
+
+    const changeModelName = (event: any) => {
+        setSelectedType(event.target.value);
```
SferaDev: In ``development-malaria`` we now have ``onValueChange`` like the ``Toggle`` component and already returns a ``string``,

.eslintrc.js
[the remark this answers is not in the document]
xurxodev: ok thanks!

src/logic/synchronization.ts
xurxodev: Done

src/models/syncRule.ts
xurxodev: done

src/types/synchronization.d.ts
xurxodev: done!

src/components/rules-creation-page/SyncRulesWizard.jsx
xurxodev: Done!

src/models/syncRule.ts
xurxodev: Done!

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
xurxodev: Done!

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
xurxodev: The development branch has material-ui 3 and makeStyles does not exists in this version.

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
xurxodev: Done!

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
xurxodev: Done!

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
SferaDev: Okay then. we could use the HOC and migrate with malaria to the hook version.

```
export default withStyles(styles)(Component name);
```

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
```diff
@@ -0,0 +1,234 @@
...
+            oldIncludeRules.indexOf(rule)
+        );
+        const ruleIndexesToInclude = _.difference(includeRules, oldIncludeRules).map(rule =>
+            oldExcludeRules.indexOf(rule)
+        );
+
+        if (ruleIndexesToInclude.length > 0) {
+            onChange(syncRule.moveRuleFromExcludeToInclude(type, ruleIndexesToInclude));
+        } else if (ruleIndexesToExclude.length > 0) {
+            onChange(syncRule.moveRuleFromIncludeToExclude(type, ruleIndexesToExclude));
```
SferaDev: [food for thought] I'm not sure sending out the indexes to the model is the best approach here. I would've just sent the difference'd arrays instead and in the method use the rule instead of its index.

This could be a big change, so it's okay to me leave it as-is for now. But having to map-to-index to later undo the map with an includes(index) seems an overkill. Since we're sending the reference in function calls, the cost of sending the whole array or the indexes array is the same.

src/utils/extractParentsFromRule.tsx
```diff
@@ -0,0 +1,18 @@
+import _ from "lodash";
+
```
SferaDev: Probable rename as ``extractChildrenFromRule`` and the file could be something like ``utils/metadataIncludeExclude.ts`` (note that this is not a react file so it shouldn't be TSX.

```
const childrenFromRule = (rule: string) =>
                oldIncludeRules.filter((children: string) => children.includes(rule + "."));
```

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
xurxodev: Done!

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
xurxodev: Done!

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
xurxodev: Done! Thanks for the explanation

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
xurxodev: Done!

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
xurxodev: Done!

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
xurxodev: ok!

src/models/syncRule.ts
```diff
@@ -133,11 +153,146 @@ export default class SyncRule {
...
+            const excludeIncludeRules = {
+                includeRules: oldIncludeRules.filter(
+                    rule => !rulesToExcludeWithChildren.includes(rule)
+                ),
+                excludeRules: [...oldExcludeRules, ...rulesToExcludeWithChildren],
+            };
+
+            return this.updateIncludeExcludeRules(type, excludeIncludeRules);
+        }
+    }
```
SferaDev: Following https://github.com/EyeSeeTea/metadata-synchronization/pull/323/files#r371679285

If we took the rules instead of its indexes we could remove the ``includes(index)``. Also I believe that we do not even need the ``rulesToInclude`` and ``rulesToExclude`` filter as these should arrive as parameters now.

Helper method ``childrenFromRule`` could be moved to utils same as ``extractParentsFromRule``.

Sintax improvement:
```
const { includeRules: oldIncludeRules, excludeRules: oldExcludeRules } = this.metadataExcludeIncludeRules[type];
```

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
xurxodev: Initially, I sent the strings of the rules but, what happens if they the sync rule receive a string of a rule that does not exist? throw an error? should the model allow this possible error? for this reason, use the indexes I think it is less prone to bugs

SferaDev:
@adrianq Aside from the comments, the rest LGTM, there're some changes that will need to be hold-back to the malaria change due to old dependencies in this branch. We will take care of them on the next PR.

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
SferaDev: Well actually with indexes we would face a worse behavior error. If it does not exist we would be selecting a different rule altogether and we wouldn't notice because there's no error handling.

Since we are building the list of possible rules, we should assume as a pre-condition that we send back correct strings that exist. Also using indexes makes the code/logic/debugging more cryptic.

src/utils/extractParentsFromRule.tsx
xurxodev: I did not move childrenFromRule to utils because oldIncludeRules is a variable of moveIncludeExcludeRules method in SyncRule, then seems not very util this function in a separate file under utils folder.

SferaDev: Understood, yeah I meant increasing the parameters ``(allRules: string[], rule: string)``. But it's your call! :)

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
xurxodev: It is understood!! This could be a big change because it is necessary to refactor unit tests.
So, do I make this change now or when we move this feature to development-malaria?

src/models/syncRule.ts
xurxodev: So, do I make this change now or when we move this feature to development-malaria?

SferaDev: We can wait, no worries.

src/components/rules-creation-page/steps/IncludeExcludeSelectionStep.tsx
xurxodev: Done!

src/utils/extractParentsFromRule.tsx
xurxodev: Done!

adrianq:
@xurxodev Closing this as we have just merged development-malaria into development and there is no relevant work here

PR #326, opened 2020-01-29
Feature/authorization routes by user role

### LINKED ISSUE #291: Add validation for user permission when the user introduces the url in the browser

xurxodev:
### :pushpin: References

* **Issue:** Closes #291

### :memo: Implementation

* I have created a new Authorization component that wrapped the component to render in the route. Authorization component must receive the function to know if the user is authorized for the route.
* If a user is authorized to show the children components
* if a user is not authorized to show an unauthorized message
* I have used current shouldShowDeletedObjects method in permissions file for route /sync/deleted
* I have added a new method verifyUserHasAccessToSyncRule in permissions file for route /sync-rules/:type(metadata|aggregated|events)/:action(new|edit)/:id?

### :art: Screenshots

Captura de pantalla 2020-01-29 a las 9 27 19

**Use case1** - login in dhis2 with a user with METADATA_SYNC_SHOW_DELETED_OBJECTS role then to navigate to http://localhost:8081/#/sync/deleted should works
**Use case2** - login in dhis2 with a user without METADATA_SYNC_SHOW_DELETED_OBJECTS role then to navigate to http://localhost:8081/#/sync/deleted should show an unauthorized text
**Use case3** - login in dhis2 with a non-admin user and try to enter in an aggregated 
 sync rule with disabled write permission for all users typing the URL or realizing double click from the list then the app should show an unauthorized text
**Use case4** - login in dhis2 with an admin user and try to enter in an aggregated 
 sync rule with disabled write permission for all users typing the URL or realizing double click from the list then the app should show the page 
**Use case5** - login in dhis2 with a non-admin user with METADATA_SYNC_CONFIGURATOR role and try to enter in an aggregated sync rule with enabled write permission for all users typing the URL or realizing double click from the list then the app should show the page.
**Use case6** - login in dhis2 with a non-admin user and try to enter in a events 
 sync rule with disabled write permission for all users typing the URL or realizing double click from the list then the app should show an unauthorized text
**Use case7** - login in dhis2 with an admin user and try to enter in a events 
 sync rule with disabled write permission for all users typing the URL or realizing double click from the list then the app should show the page 
**Use case8** - login in dhis2 with a non-admin user with METADATA_SYNC_CONFIGURATOR role and try to enter in a events sync rule with enabled write permission for all users typing the URL or realizing double click from the list then the app should show the page 
**Use case9** - login in dhis2 with a non-admin user and try to enter in a metadata 
 sync rule with disabled write permission for all users typing the URL or realizing double click from the list then the app should show an unauthorized text
**Use case10** - login in dhis2 with an admin user and try enter in a metadata 
 sync rule with disabled write permission for all users typing the URL or realizing double click from the list then the app should show the page 
**Use case11** - login in dhis2 with a non-admin user with METADATA_SYNC_CONFIGURATOR role and try to enter in a metadata sync rule with enabled write permission for all users typing the URL or realizing double click from the list then the app should show the page 

### :bookmark_tabs: Others

Bug found: Share settings by user does not work. I have tested this issue disabled write permissions of a rule for all users but I could not test the disable by specific user because not working.

Bug found: Deleted object link only is showed in landing page if the user has role METADATA_SYNC_SHOW_DELETED_OBJECTS. For an Admin user without this role the link is hidden and I think for admin the deleted object link should always be visible.

I have duplicated logic realized in SyncRulesListPage component that to decide if a user has permission to edit a sync rule and I have added a new method verifyUserHasAccessToSyncRule in a permissions file.
I have not refactored current logic in SyncRulesListPage because I think is better to talk in a web team meeting.
I recommend avoiding writing application logic like this (authorization) in the presentation layer to avoid such situations if this logic it's necessary for several components. 
In architectures such as clean architecture or hexagonal architecture, this type of logic (application logic) would be located in a layer of reusable services or use cases.


src/components/authorization/Authorization.tsx
```diff
@@ -0,0 +1,35 @@
+import React, { useState, useEffect } from "react";
+import i18n from "../../locales";
+import { Typography } from "@material-ui/core";
+
+interface AuthorizationProps {
+    authorize: () => Promise<boolean>;
+}
+
+const Authorization: React.FC<AuthorizationProps> = props => {
```
SferaDev: [minor] Use destructuring on method definition:

```
({ authorize, children }) => {...}
```

src/components/authorization/Authorization.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+    useEffect(() => {
+        async function executeIsAuthorize() {
+            const authorized = await props.authorize();
+
+            setIsAuthorize(authorized);
+        }
+
+        executeIsAuthorize();
+    }, [props]);
```
SferaDev: 1. Adding ``props`` to the dependency array might lead to duplicated side effects. Instead ``props.authorize`` would be the correct dependency (or just ``authorize`` if we destructure the props object).

2. A cleaner way to handle async operations in useEffect is to use ``then``: 

```
useEffect(() => {
    authorize().then(setIsAuthorize);
}, [authorize]);
```

src/utils/permissions.ts
```diff
@@ -79,6 +80,20 @@ export const isAppExecutor = async (d2: D2) => {
     return globalAdmin || !!userRoles.find((role: any) => role.name === name);
 };
 
+export const verifyUserHasAccessToSyncRule = async (d2: D2, syncRuleUId: string) => {
```
SferaDev: I believe we are not considering create action here.

SyncRuleId would be undefined when ``/sync-rules/:type/new`` and we would return true if user is ``appConfigurator``.

src/utils/permissions.ts
```diff
@@ -79,6 +80,20 @@ export const isAppExecutor = async (d2: D2) => {
...
+    const appConfigurator = await isAppConfigurator(d2);
+    const globalAdmin = await isGlobalAdmin(d2);
+    const userInfo = await getUserInfo(d2);
+
+    if (globalAdmin) return true;
```
SferaDev: If user is ``globalAdmin`` we can return before doing the other two async operations and it will "speed-up" a bit the check.

```
const globalAdmin = await isGlobalAdmin(d2);
if (globalAdmin) return true;
const appConfigurator = await isAppConfigurator(d2);
const userInfo = await getUserInfo(d2);
...
```

src/utils/permissions.ts
xurxodev: Done! New routes was working but I have modified the code to more readable

src/utils/permissions.ts
xurxodev: Done!

src/components/authorization/Authorization.tsx
xurxodev: Done!

src/components/authorization/Authorization.tsx
xurxodev: Done!

adrianq:
@xurxodev I have tested the PR and seems to be working like a charm.

Regarding this comment:

> Bug found: Share settings by user does not work. I have tested this issue disabled write permissions of a rule for all users but I could not test the disable by specific user because not working.

I couldn't reproduce the error. I will give you a shout tomorrow morning so we can try to reproduce it together.

@SferaDev I noticed there is a problem with the default left-click action on sync rules. As you know, in the sync rules pages the default action is the 'edit'. However, if the user doesn't have the permissions (he is not a METADATA_SYNC_CONFIGURATOR) we should change it for 'See Details'. Otherwise, it is particularly strange to be sent to the edit wizard and get immediately an unauthorized error. Can you please have a look at this tomorrow? I guess it should be quick

SferaDev:
> @SferaDev I noticed there is a problem with the default left-click action on sync rules. As you know, in the sync rules pages the default action is the 'edit'. However, if the user doesn't have the permissions (he is not a METADATA_SYNC_CONFIGURATOR) we should change it for 'See Details'. Otherwise, it is particularly strange to be sent to the edit wizard and get immediately an unauthorized error. Can you please have a look at this tomorrow? I guess it should be quick

Fixed in #331 as it does not make sense to fix it in the old table (see https://github.com/EyeSeeTea/metadata-synchronization/pull/331/commits/8181c94518d506c918210ed3447d23f5ed6a6a3a)

PR #329, opened 2020-02-03
Feature/bring include exclude selection step to development malaria

### LINKED ISSUE #216: Add option to include/exclude dependencies on sync

### LINKED ISSUE #338: In summary step for metadata sync the exclude/include rules appears in raw text, no friendly names

xurxodev:
* **Issue:** Close #216 and close #338 

### :memo: Implementation

Bring development branch solution to development-malaria

- I have created the IncludeExcludeSelectionStep component
- I have created two new fields in SyncRule: useDefaultIncludeExclude and metadataIncludeExcludeRules
- The IncludeExcludeSelectionStep the first time has the use default configuration toggle to true. With this configuration, the synchronization to use exclude include rules hardcoded in d2models. (the fields in sync rule for this configuration are useDefaultIncludeExclude= true and metadataIncludeExcludeRules= {})
- When the use default configuration toggle is set to false by the user then SyncRule assigns MetadataIncludeExcludeRules like exclude include rules hardcoded in the model for every metadata type according to selected metadata in the previous step. Then the user can realized modifications in include and exclude rules for every metadata type.
- When metadataId changes IncludeExcludeSelectionStep is initialized to default configuration because is possible that rules added to exclude, include for a type has no sense because this type now does not exist in metadataIds.
- I have created friendly names to exclude include rules in a dictionary and has been created in the IncludeExcludeSelectionStep component because it’s a presentation layer concern.
- I have added IncludeExcludeSelectionStep info in the save step
- I have added logic related to parent rules. When a rule is moved from excluded to included the parents if exists, must to be moved also: for example if the user moves from exclude to include organisationUnitGroups.organisationUnitGroupSets.attributes then organisationUnitGroups.organisationUnitGroupSets and organisationUnitGroups must be moved to include because is not possible sync attributes of organisationUnitGroupSets if this organisationUnitGroupSets does not exist in target instance for example.
- I have added logic related to children rules. When a rule is moved from include to excluded the children if exists, must be moved also: for example if the user moves from include to exclude organisationUnitGroupSets then organisationUnitGroups.organisationUnitGroupSets and organisationUnitGroups.organisationUnitGroupSets.attributes must be moved to avoid the explained problems in the previous point.
 
To found parents and children I have not created extra arrays or dictionaries, I have used the current exclude include rules in models. 

For example:
Rule: organisationUnitGroups.organisationUnitGroupSets.attributes
Parents: [organisationUnitGroups.organisationUnitGroupSets, organisationUnitGroups]

I have created the extractParentsFromRule function to this requirement and tests to validate it.

Rule: organisationUnitGroups
Children: either rule in its model that contains “organisationUnitGroups.”

### :art: Screenshots

include-exclude-step
save-step

**Use case 1**:  create a new sync rule and use default configuration toggle in the include exclude selection step should have marked by default 
**Use case 2**:  When the toggle is marked to false, then for every type of metadataIds should be assigned the default exclude include rules hardcoded in the models and the user should can realize changes.
**Use case 3**:  When metadataIds are selected and exclude include rules also, if we go back to metadata step and realize changes, then use default configuration toggle in exclude include selection step should be initialized to true.
**Use Case 4**: Navigate for steps in existed metadata sync rule with custom exclude include rules should not mark as true the toogle
**Use case 5**:  in exclude include selection step when a rule is moved from exclude to include, parents of the rule, if exists, should be moved automatically also.
**Use case 6**:  in exclude include selection step when a rule is moved from exclude to include and its parent exists previously in the include list should not create duplicates in the include list.
**Use case 7**:  in exclude include selection step when a rule and parents are selected and moved from exclude to include, should not exists duplicates in the include list.
**Use case 8**:  in exclude include selection step when a rule is moved from include to exclude, children of the rule, if exists, should be moved automatically also.
**Use case 9**:  in exclude include selection step when a rule and children are selected and moved from include to exclude, should not exists duplicates in the exclude list.
**Use case 10**:  when a sync rule is executed with use default configuration toggle in exclude include selection step mark to true then should to synchronize only relationships in include list for every model.
**Use case 11**:  when a sync rule is executed with use default configuration toggle in exclude include selection step mark to false then should to synchronize using hardcoded exclude include list in every model class.
**Use case 12**:  when a sync rule contains use default configuration toggle in the exclude include selection step mark to true then save step should contain only toggle information.
**Use case 13**:  when a sync rule contains use default configuration toggle in the exclude include selection step mark to false then save step should contain the toggle information and exclude include rules for every selected model type in metadata step.
**Use case 14**:  OnDemand synchronizations should work the same way before

### :bookmark_tabs: Others

This PR contains a lot of logic, this logic has been created in the model SyncRule to have isolated the logic, this does possible unit tests creation.

To create the logic of this issue have used TDD, as a consequence, many new tests have been added while the code was created. The use of TDD is slower creating the solution the first time but I hope on the future these tests simplify the future refactors and modifications.

Many use cases to test to verify this PR are testing using unit testing, if you found some case that is not implemented as a unit test, please let me know.

**Note**: SyncRule model is a class with many responsibilities and getting bigger and bigger
, in the future when we have time, a refactor to split the responsibilities may be a good option, but I think that I need to know better your development architecture to propose improvements.


xurxodev:
@adrianq 
 - I think that one change realized by @SferaDev, now introduces an error that we had already spoken in the PR of the development branch, this change:
https://github.com/EyeSeeTea/metadata-synchronization/pull/323#discussion_r370500615
With this change, I think the Use case 4 now is not working.

- now the lodash changes are more readable for you and less readable for me :)

@adrianq  For the rest, I don't know if all use cases described above work with these changes.

SferaDev:
> * I think that one change realized by @SferaDev, now introduces an error that we had already spoken in the PR of the development branch, this change:
>   [#323 (comment)](https://github.com/EyeSeeTea/metadata-synchronization/pull/323#discussion_r370500615)

I believe I did not see that comment, in any case ``updateMetadataIds`` is not called when the user navigates between steps. It's only called when the table receives an update (see [MetadataSelectionStep.tsx#L73](https://github.com/EyeSeeTea/metadata-synchronization/blob/feature/bring_include_exclude_selection_step_to_development_malaria/src/components/sync-wizard/common/MetadataSelectionStep.tsx#L73)).

If you could please provide an scenario where the side effect happens, we can revert that commit (although it would be the only place in the app we redundantly check for ``isEquals`` on a setter).

> For the rest, I don't know if all use cases described above work with these changes.

I've tested the changes and the behavior is the same and your jest tests pass as expected (only the order has changed now and I've updated the tests to reflect that children/parent go before rules). However further testing should be done during tomorrow to completely verify the PR is good to go.

Issue #276, opened 2019-12-26
Switch old objects table to new ui components in deleted objects section


SferaDev:
- [x] Instance configuration
- [x] Sync Rules List Page
    - [x] DRY download JSON code
- [x] Deleted objects

PR #331, opened 2020-02-03
[Feature] Update sync rules list page to new table

SferaDev:
### :pushpin: References

* **Issue:** Work in #276 

### :memo: Implementation

- Migrate ``SyncRulesList`` page to TS and the new ObjectsTable
- Fix merge from ``development`` regarding errors in ``SyncSummary``
- Switch default action in ``SyncRulesList`` from edit to details depending if the user has permissions.
- Fix bug with excluded ids being not cleared when parent was selected
- Fix bug counting invalid number of elements selected with children subtree
- Fix mapping bug with default COC (completely different instances)

- DeletedObjects is still pending to finish #276 (not a priority?)


xurxodev:
@SferaDev I have tested with this branch and I have found two bugs:
* With the new table, now the pointer cursor is never assigned for the rows. I think if the user can edit the sync rule then the pointer cursor should be visible when the cursor is over the sync rule.
* Edit action is disabled as the primary action only when the user has not METADATA_SYNC_CONFIGURATOR role but if the user can not edit the sync rule by share settings the edit action is the primary action and edit action should not be the primary action in this case.

PR #336, opened 2020-02-04
[Next] Update dependencies to latest version

SferaDev:
### :memo: Implementation

- Update dependencies to latest version
- Ignore ``cryptr`` because upgrading makes passwords to be invalid
- Add ``rxjs-compat`` and ``styled-jsx`` as they're now required as peer-dependencies
- Update ``SyncParamsSelector`` to match current style
- Add some missing exclude rules to ``DataElementModel``


xurxodev:
@SferaDev @adrianq 
The new exclude rules in data element model don't exist for current Sync rules with customed exclude and include rules. as it does not exist a production version, it is not a problem. If there were a production version we should do a migration in this scenario.

SferaDev:
Yeah I thought of this, right now we have a conceptual incoherence in this feature.

Originally excludeRules were properties that should be removed from the payload (ignored). While includeRules were REFERENCE properties that should be included into the payload.

With our current implementation we don't follow those definitions anymore, and we have overly simplified the concepts to a single MultiSelector. 

The total options of the MultiSelector could be potentially calculated from the schema without any problem (this should avoid dataStore migrations too). So the excludeRules wouldn't be used anymore in the MultiSelector (as they never were the same concept as ``!includeRules``).

However it would still be useful to allow deciding which optional properties (which can be default and or empty arrays) I want to remove from the payload. Maybe with another MultiSelector but not sure how. 

Since what we have right now should be enough for the clients that requested the feature and these are big changes, I omitted these from my review.

tokland:
I used version in this PR for project-monitoring-app, worked fine, just had to update those dependencies to avoid warnings on yarn install:

```
--- package.json
+++ package.json
@@ -80,6 +80,7 @@
         "@types/react-dom": "^16.9.5",
         "@types/react-router-dom": "^5.1.3",
         "@typescript-eslint/eslint-plugin": "^2.19.0",
+        "@typescript-eslint/parser": "^2.0.0",
         "babel-core": "^6.26.3",
         "babel-eslint": "^10.0.3",
         "cypress": "^4.0.1",
@@ -91,11 +92,11 @@
         "eslint-config-prettier": "^6.10.0",
         "eslint-config-react-app": "^5.2.0",
         "eslint-plugin-cypress": "^2.9.0",
-        "eslint-plugin-flowtype": "^4.6.0",
+        "eslint-plugin-flowtype": "^3.13.0",
         "eslint-plugin-import": "^2.20.1",
         "eslint-plugin-jsx-a11y": "^6.2.3",
         "eslint-plugin-react": "^7.18.3",
-        "eslint-plugin-react-hooks": "^2.3.0",
+        "eslint-plugin-react-hooks": "^1.7.0",
```

adrianq:
@SferaDev  I have applied @tokland suggestions and we have got rid of a few warnings. Everything seems to be working. Just an unrelated bug: https://github.com/EyeSeeTea/metadata-synchronization/issues/375

PR #348, opened 2020-02-06
[Feature] Add disabled state and fix bug with summary

### LINKED ISSUE #344: Add 'no mapping' state
Include 'no mapping' state to the mapping elements. By default, all objects are automatically mapped to the same uid as now. However, we should include a new contextual action 'Clear Mapping' that delete even the automatic mapping. We will end up with two contextual actions: 'Reset to default values' (that will perform the current behaviour we have in clear mapping) and 'Clear mapping' that will set the object to no mapped.  No mapped objects will no be synced even if they are part of a sync rule. No mapped objects should be stored in the data store (probably something like what we have for excluded objects). We will not store automatic mapping as we currently do (no changes here).
- [x] Implement no mapping state and store in data store.
- [x] Exclude no mapped object from sync process
- [ ] Add contextual actions: 'Reset to default values' and 'Clear Mapping'
- [ ] Add two new buttons on top right corner 'Reset to default values' and 'Clear Mapping'. These buttons will execute this action on all available elements in the table. A confirmation dialog will be displayed before performing the action

SferaDev:
### :pushpin: References

* **Issue:** Closes #344 

### :memo: Implementation

-

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2020-02-06T20:51:02.363Z\n"
-"PO-Revision-Date: 2020-02-06T20:51:02.363Z\n"
+"POT-Creation-Date: 2020-02-06T21:04:27.739Z\n"
+"PO-Revision-Date: 2020-02-06T21:04:27.739Z\n"
 
 msgid "Unauthorized - You do not have permission to view this page."
 msgstr ""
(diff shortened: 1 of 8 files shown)


xurxodev:
@SferaDev @adrianq 
I have found a situation that I don't know if is an error or not but I take me many time found the problem and I think is not related to this PR:

I have tested with aggregate sync with http://taris.sferadev.com:9006/ as the sender and http://taris.sferadev.com:9007 as the receiver. I have tested the 'Jorge test' data set where I have a data element with category combination and this category combination is duplicated in the receiver and I have realized mapping with wrong duplicated category combination (not assigned to data element in receiver).

With this configuration the synchronization show success in summary but this values was empty in the receiver data entry. When I have found the problem and I have change de mapping between category combinations to map to correct category combination assigned in data element in the receiver, the sync works successfully.

Should we show error in this case in summary, can we do it?

SferaDev:
We could try to do an intelligent warning system by checking the dataElement but this is not an error in DHIS2 because a dataSet has the capability to override the dataElement category.

PR #355, opened 2020-02-10
[Feature] Add dry run option

### LINKED ISSUE #328: perform manual check: 'Validate'
To be discussed with @adrianq 

- [x]  Auto-map suggestion: Contextual action that queries remote instance for a DE/OU/COC/... that matches code/name (ilike, startsWith, token) and opens the mapping dialog with the filters applied to the ones that match => Moved to #359
- [x] View the metadata mapping that will be applied in the summary for data synchronizations (on-demand and sync rule) => @xurxodev => Move to #370
- [x] Add an option to execute the synchronization without committing changes (dry run) => @xurxodev => Move to #357

Missing tasks from https://github.com/EyeSeeTea/metadata-synchronization/issues/161
- [x] Add button to perform manual check: 'Validate'
- [x] Modify table loading spinner to include a message (d2-ui-components). It should say something like 'Checking destination org units'

### LINKED ISSUE #357: Dry Run
Add an option to execute the synchronization without committing changes (dry run)

xurxodev:
* **Issue:** close #357 

### :memo: Implementation

- I have added toggle dry run parameter to instance selection step in SyncParamsSelector in metadata, events and aggregates
- I have added dry run info in summary step for metadata, events and aggregates
- I have modified the synchronize process to retrieve dry run param from sync rule if exists

For metadata params:
Param is import mode
Dry run on: COMMIT
Dry run off: VALIDATE

For data params:
param is dryRun

### :art: Screenshots
Captura de pantalla 2020-02-10 a las 9 12 49
Captura de pantalla 2020-02-10 a las 9 13 20

**Use case 1**:  Dry run should works for metadata, events and aggregates
**Use case 2**:  Dry run parameter should be visible in summary step for metadata, events and aggregates


src/utils/synchronization.ts
```diff
@@ -104,15 +104,14 @@ export async function postMetadata(
 ): Promise<MetadataImportResponse> {
     try {
         const params: MetadataImportParams = {
-            importMode: "COMMIT",
+            importMode: additionalParams?.importMode ?? "COMMIT",
```
SferaDev: There's no need for this setting here, as in line 113 we're already destructuring ``additionalParams``. So ``importMode: "COMMIT"`` here is the default behvaiour and if ``importMode`` is set it will overwrite the default value in the destructuring.

src/utils/synchronization.ts
```diff
@@ -428,7 +427,7 @@ export async function postData(
                     skipExistingCheck: false,
                     format: "json",
                     async: false,
-                    dryRun: false,
+                    dryRun: additionalParams?.dryRun ?? false,
```
SferaDev: There's no need for this setting here, as in line 431 we're already destructuring ``additionalParams``. So ``dryRun: false`` here is the default behvaiour and if ``dryRun`` is set it will overwrite the default value in the destructuring.

src/utils/synchronization.ts
xurxodev: Done!, 
I didn't know that the order of destructuring and assigning fields affected to be default or assigned value, thanks.

src/utils/synchronization.ts
xurxodev: Done!, 
I didn't know that the order of destructuring and assigning fields affected to be default or assigned value, thanks.

PR #358, opened 2020-02-10
[Feature] Auto-mapping feature

SferaDev:
### :pushpin: References

* **Issue:** Work in #328

### :memo: Implementation

- Add auto-mapping feature as a single row action

### :art: Screenshots

image
image
image

- I tried to implement it as a multiple row action but it leads to unwanted race conditions in the model ``save`` operation.


src/pages/instance-mapping/InstanceMapping.tsx
```diff
@@ -198,6 +199,50 @@ const InstanceMappingPage: React.FC = () => {
...
+            return;
+        }
+
+        const { objects: candidates } = await instance
+            .getApi()
+            //@ts-ignore
+            .models[type].get({
+                fields: { id: true },
+                filter: {
+                    name: { ilike: selectedItem.name },
```
adrianq: @SferaDev Not sure about the logic here. selectedItem.name is empty. Maybe we can use displayName o modify the model. Apart from that, I have some doubts regarding the API call. For example, create a new data element with the same name than in the origin instance and a completely different short name and code. The auto mapping feature cannot find the data elements. I guess the API call needs to be refined...

PR #377, opened 2020-02-12
[Testing] Add cypress test for on-demand aggregated sync

### LINKED ISSUE #375: Error escaping characters on loading message
image

SferaDev:
### :pushpin: References

- **Issue:** Closes #375 
- https://github.com/EyeSeeTea/d2-ui-components/pull/128

### :memo: Implementation

- Update d2-api to latest beta and update some deprecated imports
- Add cypress UI testing for on-demand aggregated sync


xurxodev:
I have finished to refactor this PR to use page object.

The basic idea is to have an object abstraction that represents the data and actions of the page involved in the tests: 
* Data is represented as properties in the page object and assertions from tests should be realized on the data.
* Actions are represented as normal functions and normally as fluent API, that is to say, to simplify the execution of functions from tests, every action function returns the object itself (in this sense it works like a builder).

I have added new tests for you can see the utility to use page objects and it's very easy and with readable code.

Added tests:
- different scenarios for validation errors where a SnackBar should appear.
- different scenarios where the synchronize button should be enabled or disabled.

I have used javascript to page object and test because to use TypeScript is necessary to realize changes in commands definitions and I don't want to do this PR bigger and to separate concerns. 
Look this [cypress documentation for typescript](https://docs.cypress.io/guides/tooling/typescript-support.html).

If this approach seems good to you, the next steps related to the test could be:
* Refactor landing page tests and instance configurator tests to use page objects.
* Refactor cypress code to typescript.
* Create more tests on other pages.

cypress/integration/manual-aggregated-sync.spec.js
```diff
@@ -1,76 +1,136 @@
-import { dataTest } from "../support/utils";
+import ManualAggregateSyncPageObject from "../pageobjects/ManualAggregateSyncPageObject";
 
+/**
+ * Database: d2-docker-eyeseetea-2-30-datasync-sender
+ */
 context("Manual aggregated sync", function() {
+    const page = new ManualAggregateSyncPageObject(cy);
```
SferaDev: LGTM, I would make the PageObject agnostic to "aggregated", "events" and "metadata" though and pass it here as an argument. I believe the same PageObject can be re-used for the three of them. (Similarly this would happen on SyncRules and History PageObjects)

xurxodev: It is a possibility, however, I would not like to fall into premature abstractions since they can be misleading. If the page object can be used for all manual synchronizations I think it is something to review in the future PR when we create the next on-demand tests.

SferaDev: Well, they're the same component with different models so the abstraction is already there but in the application code. I believe these PageObjects should mimic the application design and naming so it's consistent and easy to follow, anyway it's okay to leave it to a future PR.

cypress/integration/manual-aggregated-sync.spec.js
```diff
@@ -1,27 +1,137 @@
...
+
+    const anyOrgUnit = "Ghana";
+    const anyInstance = "Y5QsHDoD4I0";
+    const anyDataset = "Malaria annual data";
+
```
tokland: Minor: why are those called any? Maybe group in an object.

cypress/integration/manual-aggregated-sync.spec.js
```diff
@@ -1,27 +1,137 @@
...
+            .selectAllPeriods()
+            .next()
+
+            .selectAllAttributesCategoryOptions()
+            .next()
+
+            .selectReceiverInstance(anyInstance)
+            .synchronize()
+
+            .syncResults.contains("Success");
```
tokland: This assertion on attributes breaks the chain, does it make sense to be able to perform those checks as a chainable method?

cypress/integration/manual-aggregated-sync.spec.js
```diff
@@ -1,27 +1,137 @@
...
+
+        page.closeSyncResultsDialog();
+    });
+
+    it("should show the org unit step error if user try click on next without selecting the org unit", function() {
```
tokland: minor: from ES6 we tend to write lambdas as arrow funcions `() => {}`

package.json
```diff
@@ -61,7 +61,7 @@
     "pre-push": "yarn prettify && yarn lint && yarn localize && yarn jest",
     "cy:verify": "cypress verify",
     "cy:e2e:open": "CYPRESS_E2E=true cypress open",
-    "cy:e2e:run": "CYPRESS_E2E=true cypress run"
+    "cy:e2e:run": "CYPRESS_E2E=true cypress run --browser chrome"
```
tokland: Once we updated to latest packages on project-monitoring we no longer needed to use the chrome browser, the default electron worked fine. Just as a note, probably there is nothing wrong with using chrome, now cypress can record videos also with Chrome.

cypress/pageobjects/ManualAggregateSyncPageObject.js
```diff
@@ -0,0 +1,117 @@
+import { dataTest } from "../support/utils";
```
tokland: Minor: not sure about adding a new folder to cypress. Maybe using `cypress/suport`?

cypress/integration/manual-aggregated-sync.spec.js
tokland: I agree with both your points. On the next iteration, once we write tests for the events/metadata/..., we can see the proper abstraction more clearly.

package.json
SferaDev: I found that with electron the cookies are not separated by port, so when performing synchronizations between two instances on localhost the JSESSION is overwritten and the tests do not pass.

Right now we don't have two dockers yet but once we set them up, we will need to run them on chrome.

cypress/integration/manual-aggregated-sync.spec.js
xurxodev: There is no special reason, in Android I usually use constants and call them anyX because the value to understand the test does not matter, ok, I will to remove any keyword and create an object to group variables.

cypress/integration/manual-aggregated-sync.spec.js
xurxodev: It's true these methods break the chain. Normally this is not a problem because the validation about data usually is the last statement. But if you need chain more actions then the chain is broken. Ok, I'm giving some more thought about this.

cypress/integration/manual-aggregated-sync.spec.js
xurxodev: Ok, I will do it

cypress/integration/manual-aggregated-sync.spec.js
xurxodev: In the next iteration for the moment I'm currently opting for a base class but it is not finished

cypress/pageobjects/ManualAggregateSyncPageObject.js
xurxodev: Yes, no problem.

cypress/integration/manual-aggregated-sync.spec.js
xurxodev: Done!

cypress/integration/manual-aggregated-sync.spec.js
xurxodev: Done!

cypress/pageobjects/ManualAggregateSyncPageObject.js
xurxodev: Done!

cypress/integration/manual-aggregated-sync.spec.js
xurxodev: Done! I have replaced data properties to chainable asserts methods.
The goal is not to break the chain and leave the assert responsibility to the test

PR #381, opened 2020-02-14
Feature/view metadata mapping in the summary

### LINKED ISSUE #370: View the metadata mapping that will be applied in the summary for data synchronizations (on-demand and sync rule)

xurxodev:
### :pushpin: References

* **Issue:** Closes #370

### :memo: Implementation

- I have realized a request to retrieve target instances with metadata mapping
- I have added new node children to the summary tree to render metadata mapping for aggregates and events
- Current stored metadata mapping for an instance only contains types and uids for source and target instance, As we talked at the meeting I have not done any additional request to retrieve display names to show.
-  I have added filter for metadata mapping by existed models in metadata of the sync rule
(example: organisationUnits) we are not filtering metadata mapping by metadata related to data (aggregate, events) for example by data elements, CategoryOption, this filter will be realized on the future in other issues. Then is possible we need to use here other metadata array variable to use for filters.

I have created a new function to render metadata mapping above of the return statement because with current size and indent of the return statement, this section leaves the code unreadable. 

### :art: Screenshots

Captura de pantalla 2020-02-14 a las 9 39 20

* **use case1** - Metadata mapping should be visible for every selected target instance in the summary step for aggregates and events.

### :bookmark_tabs: Others

The summary step component is very big and to add new nodes in the tree with the current indent level leave the code unreadable. 

I would like to propose a refactor to improve it:

Currently, we have mixed data tree composition and rendering. I think that the code would be more readable if we split these two responsibilities. 

First, we to create the data as a tree and then we rendering the tree using a tree component. We could use tree component of a new library or create our own tree component in d2-ui-components repository.


adrianq:
@xurxodev I am merging this as this is a first step in the right direction but there is still work to do here. As discussed, we need to filter by the relevant metadata objects. Also, we should hide the label if there is no mapping in the sync execution

PR #438, opened 2020-03-09
Add new metadata objects

### LINKED ISSUE #415: Add new metadata objects
- [x] Dashboard
- [x] User Group
- [ ] Program -> moved to #442
- [x] Dataset

xurxodev:
### :pushpin: References

* **Issue:** Closes #415 

### :memo: Implementation

- I have added the Dashboard model to metadata models 
- I have added the UserGroup model to metadata models
- I have added the DataSet model to metadata models
- I have created new tests

#### Dashboards
The sync process fails if the dashboards contain related metadata as org units and this metadata does not exist in the target instance.

#### UserGroup
The sync process fails if the dashboards contain related metadata as org units and this metadata does not exist in the target instance.

User roles can not be synced directly with UserGroup and users because roles are embedded under UserCredentials and current data sync logic not found user roles to add it in the package to upload.
A possible solution is to add the UserRole model to metadata models and sync it previously, another is created custom sync to this scenario.

#### Dataset
I have found a bug with related indicators if related IndicatorGroupSet does not exist in the target instance even if this model is in the include rule list, this include rule not working and sync fail. This bug occurs also when you sync indicator model directly.

IndicatorGroupSet bug I think is an API problem because the api/IndicatorGroups endpoint does not return the related IndicatorGroupSet.

### :bookmark_tabs: Others

### DIFF
--- a/src/models/d2Model.ts
+++ b/src/models/d2Model.ts
@@ -557,7 +557,7 @@ export class DashboardModel extends D2Model {
         "eventReports",
         "maps",
         "reports",
-        "reportTables"
+        "reportTables",
     ];
 }
 
--- a/src/models/d2Model.ts
+++ b/src/models/d2Model.ts
@@ -307,6 +307,44 @@ export class DataSetModel extends D2Model {
     protected static fields = dataSetFields;
     protected static childrenKeys = ["dataElements"];
 
+    protected static excludeRules = [
+        "indicators.dataSets",
+        "indicators.programs",
+        "dataElements.dataSets",
+        "dataElements.dataElementGroups.dataElements",
+        "dataElements.dataElementGroups.dataElementGroupSets.dataElementGroups",
+    ];
+    protected static includeRules = [
+        "attributes",
+        "legendSets",
+        "categoryCombos",
+        "categoryCombos.attributes",
+        "categoryCombos.categoryOptionCombos",
+        "categoryCombos.categoryOptionCombos.categoryOptions",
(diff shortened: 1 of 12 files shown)


adrianq:
@xurxodev Will review the PR ASAP but can you please check if the IndicatorGroupSet bug is happening for latest DHIS2 versions and open a JIRA issue if so?

adrianq:
@tokland PR seems to be working fine. I have captured the limitation and missing features here https://github.com/EyeSeeTea/metadata-synchronization/issues/444 Will discuss them with Nacho today. Add the JIRA issue link to the card once opened

xurxodev:
@adrianq I have tested with new servers build and after testing with new servers build I have verified that metadata endpoint now returns indicatorGroupSets relationships for an indicatorGroup. Then I have modified include rules to the plural and now is indicators sync is working.
I have created this PR https://github.com/EyeSeeTea/metadata-synchronization/pull/447

Dashboard bug when exists in target previously on this servers does not fail, only Alexis servers

src/models/d2Model.ts
```diff
@@ -505,6 +543,32 @@ export class ValidationRuleGroupModel extends D2Model {
...
+    protected static includeRules = [
+        "dashboardItems",
+        "charts",
+        "eventCharts",
+        "pivotTables",
+        "eventReports",
+        "maps",
+        "reports",
+        "reportTables",
+    ];
```
SferaDev, line 561: Review this properties

PR #478, opened 2020-03-25
Feature/add generate uid data param

### LINKED ISSUE #431: Add toggle 'Create with new UID'
Set to true when syncing with the same instance

xurxodev:
### :pushpin: References

* **Issue:** Closes #431 

### :memo: Implementation

The goal create a new param GenerateNewUId only for sync rules of type events and if the target instances contain current URL, then GenerateNewUId param should be true and disabled.

- I have created a new data param
- I have modified SyncParamsSelector component to add toggle for new param if sync rule is events.
- I have modified instance selection step to assign disabled and true to new param if in target instances exist the current URL.  
- I have modified events synchronization logic to assign new UID if required

### :art: Screenshots

Captura de pantalla 2020-03-25 a las 8 06 08


src/components/sync-wizard/common/InstanceSelectionStep.tsx
```diff
@@ -13,19 +13,45 @@ export const getInstanceOptions = async (api: D2Api) => {
...
     const [selectedOptions, setSelectedOptions] = useState<string[]>(syncRule.targetInstances);
     const [instanceOptions, setInstanceOptions] = useState<{ value: string; text: string }[]>([]);
+    const [targetInstances, setTargetInstances] = useState<Instance[]>([]);
+
+    const includeCurrentUrlAndTypeIsEvents = (selectedinstanceIds: string[]) => {
+        return (
+            syncRule.type === "events" &&
+            selectedinstanceIds
+                .map(id => targetInstances.find(instance => instance.id === id)?.url)
+                .includes(api.baseUrl)
```
SferaDev: [minor] [untested]
```suggestion
            selectedinstanceIds
                .map(id => _.find(targetInstances, ["id", id])?.url)
                .includes(api.baseUrl)
```

Also this method signature has ``selectedinstanceIds: string[]``, isn't it the same as ``selectedOptions``? Is there a reason to use a block-scoped ``syncRule`` and a parameter ``selectedOptions``?

xurxodev: Yes, It's the same but I need the value to invoke includeCurrentUrlAndTypeIsEvents function while selected instances are changing also if  I use selected options at this moment has not the last user action.

Regarding using lodash, I prefer not to have to use strings because it is more error-prone

SferaDev: Actually with TS strings are type-checked and "id" in that example should be treated as a constant and the compiler normally verifies if the type includes a property called ``id``.

xurxodev: There are many scenarios where to use strings is more error-prone, another example is rename refactor using rename symbol with vs code tool

PR #483, opened 2020-03-27
Assign deletedSyncRuleLabel to notifications

### LINKED ISSUE #434: Keep history for deleted rules
This is the current behaviour. The client want to keep the sync rule name. Feasible?

xurxodev:
### :pushpin: References

* **Issue:** Closes #434

### :memo: Implementation

- For each SynRule to delete I load its SyncReports (notifications in the data store, this data is shown in the history page) and I assign deletedSyncRuleLabel field and I save it.
- In the history page if deletedSyncRuleLabel exists is show it as sync rule name

- When a sync rule is removed, its histories should show deleted sync rule name + (deleted).
- When a sync rule is removed, its details in histories should show deleted sync rule name + (deleted).

### :bookmark_tabs: Others

@adrianq @tokland @SferaDev 

I have developed this feature following the current architecture of the application. However, developing this feature I have found a good example to recommend the use of [Clean Architecture](https://blog.cleancoder.com/uncle-bob/2012/08/13/the-clean-architecture.html) also in our web projects.

Clean architecture is a wide concept, this feature gives me an opportunity to explain a single concept of Clean Architecture, the use cases.

####  Use Cases in Clean Architecture

0_JD606Sqx6RYZLKdu

The software in this layer contains application-specific business rules. It encapsulates and implements all of the use cases of the system. These use cases orchestrate the flow of data to and from the entities, and direct those entities to use their enterprise-wide business rules to achieve the goals of the use case.

We also do not expect this layer to be affected by changes to externalities such as the database, the UI, or any of the common frameworks. This layer is isolated from such concerns.

Use cases should depend on abstractions (e.g. interfaces) allowing for safer future modifications

A use case is a piece of code independent of UI (React), APIs or Database (Dhis2), given that this piece of code should be isolated of libraries and frameworks like React or Dhis2.

Creating use cases we have several benefits:
* We have a list or index of the capacities of our application
* Reusable code, if there are more of one place in the application that to realize the action, the code is not duplicate.
* The code is safer because this application logic does not be affected by changes in UI or network components for example.
* if use cases depend on abstractions are more scalable to future changes.
*  if use cases depend on abstractions, we can test the logic using jest unit test using test doubles. Unit tests are faster and safer than cypress test as we are seeing.

#### Delete Sync Rule Use Case

In this feature, there are actions to realize when a Sync Rules are deleted
* To realize an iteration for each sync rule:
* for each sync rule
    * Invoke delete action to the data source (using dhis2 API)
    * Load children notifications from the data source (using dhis2 API)
    * Modify children notifications
    * Invoke save action children notifications to the data source (using dhis2 API)

This code currently is in SyncRulesPage that is a React Component.

We could create a new DeleteSyncRuleUseCase file under a use case or domain folder with these piece of code.

This is a proposal to use Clean Architecture in our web applications, I know that these type of proposals are better in meetings but I did not want to miss the opportunity to explain the concept of the use case of Clean Architecture with a specific example.

We can extend this conversation in future web meetings.

I take this opportunity to list again the resources that I have sent you on some occasions about clean architecture in the frontend.
* https://www.youtube.com/watch?v=vxx-UEpnNaY
* https://janithl.github.io/2019/10/react-clean-architecture-part-1/
* https://janithl.github.io/2019/10/react-clean-architecture-part-2/


SferaDev:
Hi @xurxodev 

Thanks for the article and example, however I believe we are in some extent already doing similar concepts to what you suggest (obviously we do not have an abstraction for DHIS2 or a real business layer but we do mostly separate logic from UI).

IMO you implemented the logic in the wrong place, I would have personally done it the following way without ever thouching the SyncRulesListPage.tsx component (which aligns more to what you suggest in the PR comment).

src/models/syncRule.ts:
```
    public async remove(api: D2Api): Promise<void> {
        await SyncReport.updateReportsOnRuleDelete(api, this);
        await deleteData(api, dataStoreKey, this.syncRule);
        const detailsKey = SyncRule.getDetailsKey(this.syncRule);
        await deleteDataStore(api, detailsKey);
    }
```

src/models/syncReport.ts:
```
    public static async updateReportsOnRuleDelete(api: D2Api, rule: SyncRule): Promise<void> {
        const deletedRuleLabel = `${rule.name} (${i18n.t("deleted")})`;

        const syncReports = await SyncReport.list(
            api,
            { type: rule.type, syncRuleFilter: rule.id },
            {},
            false
        );

        for (const syncReportData of syncReports.rows) {
            const editedSyncReport = {
                ...syncReportData,
                deletedSyncRuleLabel: deletedRuleLabel,
            };
            const syncReport = SyncReport.build(editedSyncReport);
            const syncResults = await syncReport.loadSyncResults(api);
            syncReport.addSyncResult(syncResults[0]);

            await syncReport.save(api);
        }
    }
```

NOTE: I've only copy-pasted the code from this PR, didn't check if the logic is accurate or if my snippets correctly work. It's only an example for how I would organize the code.

EyeSeeTeaBotTest:
<h3>Tests ran successfully :tada:</h3><br />Build commit Merge bd1a487128d4c70904ce6abbe5166e5d5ff7d7fa into f5a9b7b90acf252d77f6d807bb9bde54d6310bb1 with commit id b70b73d2636c3e693b11d9b359d2a5e76aad88a4 on branch feature/keep_history_for_deleted_sync_rules<br /><br />Travis log available at <a href=https://travis-ci.org/EyeSeeTea/metadata-synchronization/builds/667625952 target=_blank>builds/667625952</a><br /><br />See cypress results at <a href=https://eyeseetea.github.io/feedback-ci/EyeSeeTea/metadata-synchronization/pr/483/b70b73d2636c3e693b11d9b359d2a5e76aad88a4 target=_blank>feedback-ci/EyeSeeTea/metadata-synchronization/pr/483/b70b73d2636c3e693b11d9b359d2a5e76aad88a4</a>

xurxodev:
Hi @SferaDev 

Ok, I think we should talk about this on future web meetings.

I have followed the same line in the current delete action, loop for each SyncRule and delete method invocation for sync rule was realized in the react component and also is business logic. 

I think your solution breaks the logic for the use case and move the problem from react component to SyncReport model, but the problems are similar. Business logic coupled to data logic and frameworks or libraries and this to do SyncReport model bigger. 

Currently, our models have many responsibilities:
* Methods to retrieve data of data source (using dhis2 API)
* data logic, for example, to create datastore keys
* contains the data of the dhis 2 models
* enterprise business logic, for example realizing validations
* and your solution add partial aplication logic of a use case that its application logic.

Clean architecture would break down our models in small and reuse pieces of code.

As I say this gives for a long conversation and I think which is better in some meeting or several after the release.

tokland:
@xurxodev Thanks for the explanation! I also think we should give this some time at a team meeting. 

I guess we all agree on this: we tend to have pretty good isolation from the UI, but we are tightly coupled with the DHIS2 API. We feared that adding a layer would add complexity and interface friction. That's not without consequences, of course, we all have experienced a hard time writing unit tests.

@adrianq / @xurxodev I think it would be great to have some playground to see, with real code, how Clean Architecture would be applied to our typical React+Dhis2 apps. Something very minimal, a dhis2-skeleton-app with a component that changes the current user name or something like that, maybe?

SferaDev:
Since this is a closed PR, maybe we should mirror all the comments here and use the [Issue tracker on know-how](https://github.com/EyeSeeTea/know-how/issues) for this kind of discussions?

> I guess we all agree on this: we tend to have pretty good isolation from the UI, but we are tightly coupled with the DHIS2 API.

Yeah being tightly coupled with DHIS2 API is not easy to solve due the nature of the applications we develop. The effort to build that abstraction should be due a strategic decision (ie. being capable of provide the same functionality for other software backends) rather than application architecture.

However I believe @xurxodev also pointed that our models have many responsibilities and that some of them should belong to an upper layer (business) instead. I'm afraid that would lead to a business layer with many responsibilities and models with almost only getters/setters (in some languages those don't even exist in ``data-classes``).

> It would be great to have some playground to see, with real code, how Clean Architecture would be applied to our typical React+Dhis2 apps. Something very minimal, a dhis2-skeleton-app with a component that changes the current user name or something like that, maybe?

Maybe my comments above would change with a real-world example. That makes a lot of sense to me, +1!

PR #490, opened 2020-03-31
[Fix] Include more metadata types for MAL

SferaDev:
### :memo: Implementation

- Add ``uniqBy`` for metadata packages
- Add ``sections`` to DataSetModel
- Add ``dataEntryForms`` to DataSetModel
- Add ``indicatorTypes`` and ``legendSets`` to indicators from IndicatorGroupModel and IndicatorGroupSetModel

With these changes applied we have a very similar result (selecting the same metadata both in a SyncRule and the Advanced-Export).

These are the pending differences:

- Categories: We include them and Advanced-Export does not: this is a known issue in that app (see https://github.com/EyeSeeTea/Advanced-Metadata-Export/pull/16)
- DataElements: We miss 6 that Advanced-Export includes, all these 6 elements are marked as OLD in the name. They're not included in the DataSets anymore but it still looks like are stored in the DataSetSection (where the Advanced-Export finds them). I believe our behavior of excluding them is the correct approach.
- CategoryOptionGroups: We do not include them in MDSync but Advanced-Export includes them. They're coming from a relationship through CategoryOption. For me this is a bug of Advanced-Export exclude rules.

Improvements noticed:

- We have a limitation (and IIRC the Advanced-Export too) that if an Indicator uses a DataElement as part of a numerator or denominator we do not include it as dependency because it is not part of the metadata (only expression). Some of the Indicators have references to DataElements that are not part of the payload.

### :bookmark_tabs: Others

- This changes were not possible through the web interface and to make them possible it is required to change how the include/exclude UI feature works.
- I recall saying this somewhere in the original PRs revision but exclude is not the opposite as include. This was a misconception and we should solve it in the future.

- Include rules: Metadata elements (parsed by id) that should be recursively included in the package
    - Default values: From the model
    - All values: Calculated from schemas (through REFERENCE type)
- Exclude rules: Properties (not only references) that should be removed from the payload
    - Default values: From the model
    - All values: Calculated from schemas (through mandatory false)


EyeSeeTeaBotTest:
<h3>Tests ran successfully :tada:</h3><br />Build commit Merge 39639e7abf0b484806a0d62fd6495501d9acf2a6 into 4fb49c53b9ab88d135c54bb49e0563250f518c77 with commit id b70e94f16e461ef122bf0cbf60ae23f5796f8db3 on branch fix/ryan-packages<br /><br />Travis log available at <a href=https://travis-ci.org/EyeSeeTea/metadata-synchronization/builds/669165481 target=_blank>builds/669165481</a><br /><br />See cypress results at <a href=https://eyeseetea.github.io/feedback-ci/EyeSeeTea/metadata-synchronization/pr/490/b70e94f16e461ef122bf0cbf60ae23f5796f8db3 target=_blank>feedback-ci/EyeSeeTea/metadata-synchronization/pr/490/b70e94f16e461ef122bf0cbf60ae23f5796f8db3</a>

PR #492, opened 2020-04-02
[Feature] Group rows in global mapping for category options and options

### LINKED ISSUE #487: Group CO and options by categories and option sets in global mapping section

SferaDev:
### :pushpin: References

* **Issue:** Closes #487

### :memo: Implementation

-


adrianq:
@SferaDev It seems like now new maps are not set as global:
image
0-4 and 15+ were already mapped and I just mapped 5-14 years and it doesn't show the 'Global' label. Same happens for options/option sets.

PR #497, opened 2020-04-04
[Feature] Integrate scheduler server with application

### LINKED ISSUE #440: Review scheduler

### LINKED ISSUE #493: Migrate scheduler code to application

SferaDev:
### :pushpin: References

* **Issue:** Closes #440
* **Issue:** Closes #493


src/migrations/utils.ts
```diff
@@ -14,8 +13,7 @@ export function getDuplicatedIds<Obj extends Ref>(objects: Obj[]): string[] {
 }
 
 export function getMigrationsForNode(): Migration[] {
-    const directory = path.join(__dirname, "tasks");
-    const keys = _.sortBy(fs.readdirSync(directory));
+    const keys = _.sortBy(fs.readdirSync("./src/migrations/tasks"));
```
SferaDev, line 16: @tokland Relative path on static code analysis doesn't work and it broke the functionality to merge all the code into a single distributable file.

tokland: What do you propose? Originally, I had explicit imports, but I wanted to avoid the hassle both add the file and add the import + some array.

SferaDev: > What do you propose? Originally, I had explicit imports, but I wanted to avoid the hassle both add the file and add the import + some array.

Actually the snipped I changed (relative from project root) seems to work just fine, however I wanted your feedback on this change.

tokland: Oh, ok, I misunderstood :+1:

package.json
```diff
@@ -46,11 +48,13 @@
     "styled-jsx": "3.2.5"
   },
   "scripts": {
+    "clean": "npx rimraf build/ dist/",
     "prestart": "yarn localize && d2-manifest package.json manifest.webapp",
     "start": "react-scripts start",
     "prebuild": "yarn test",
     "build": "rm -rf build/ && yarn localize && d2-manifest package.json manifest.webapp && react-scripts build && yarn run manifest && cp -r i18n icon.png build",
     "build-webapp": "yarn build && rm -f $npm_package_name.zip && cd build && zip -r ../$npm_package_name.zip *",
+    "build-scheduler": "npx ncc build src/scheduler/cli.ts -m",
```
tokland, line 57: [very minor] Those two npx in clean/build-scheduler are necessary? It should use the ones in node_modules/.bin by default.

SferaDev: That's true, I copy/pasted my commands from the terminal

PR #510, opened 2020-04-24
Integrate new models in metadata

### LINKED ISSUE #509: Integrate new models in metadata
- Complex (10) 5-8h
- [x] Chart
- [x] EventChart
- [x] EventReport
- [x] Map
- [x] MapView
- [x] ProgramRuleAction
- [x] ProgramStage
- [x] ProgramStageSection
- [x] Report
- [x] User

- Medium (8) 3-5h
- [x] DataEntryForm
- [ ] ~~JobConfiguration~~ For the moment we don't add it
- [x] LegendSet
- [x] Section
- [x] TrackedEntityAttribute
- [x] TrackedEntityType
- [x] UserRole

- Simple (9)  2-3h
- [x] Category
- [x] CategoryCombo
- [x] CategoryOption
- [x] CategoryOptionCombo -
- [x] CategoryOptionGroup
- [x] CategoryOptionGroupSet
- [x] Option
- [x] OptionGroup
- [x] OptionSet
- [x] IndicatorType

xurxodev:
### :pushpin: References

* **Issue:** Closes #509

#### Complex (10) 
- [x] Chart 
- [x] EventChart 
- [x] EventReport
- [x] Map
- [x] MapView 
- [x] Report (embedded dependencies in designContent HTML field)

**For every Dashboard related metadata models (Chart, EventChart etc..) and Report model, I have followed the same approach with Dashboard model, I have not to include its related metadata (program, org unit, data element, categories) as rules. If this metadata does not exist on the receiver to view the item on receiver fail**.

- [x] ProgramRuleAction 

~~Program Rule Action is the first added metadata model that It has not field name then search by name has no sense, I have modified default search by name to by identifiable.
According to dhis2 doc identifiable search by id, code, and name (also shortName if available)
https://docs.dhis2.org/master/en/developer/html/webapi_metadata_object_filter.html~~

After making many tests, the conclusion is that filter by identifiable object and token is a new feature of version 2.31 as I saw here https://community.dhis2.org/t/dhis-version-2-31-is-released/34292, giving backwards compatibility they put it in 2.30 and 2.29.

During the testing I saw that it search by identifiable fails but only in 2.30, from 2.31 which is when they put it in, it works well.
The error that occurs in 2.30 is that if you add more filters after filtering by identifiable with the token operator, directly the API ignores other filters.

Example request

Request to dataElements in 2.30 filtering by identifiable
https://play.dhis2.org/2.30/api/dataElements?filter=identifiable:token:art

Then if I add a second filter the results count should be lower but the results are the same
https://play.dhis2.org/2.30/api/dataElements?filter=identifiable:token:art&filter=valueType:eq:NUMBER

Request to dataElements in 2.33 filtering by identifiable
https://play.dhis2.org/2.33/api/dataElements?filter=identifiable:token:art

Here in 2.33 if I add a second filter, works successfully and the results count are lower
https://play.dhis2.org/2.33/api/dataElements?filter=identifiable:token:art&filter=valueType:eq:NUMBER

So since we are with 2.30 with metadasync, which is where API filtering by identifiable with more filters fails, we can not use it. I have refactored the code to define by dhis2 model configurable search field and operator. By default, the field is name and operator is token but for programRuleAction the field is id and operator is ilike.

- [x] ProgramStage

**A program stage has no sense without a program. As we have the program model to sync also with all dependencies, I have included the program stage dependencies as include rules and program but without dependencies, because we need to update the program to add new program stage. It's a simpler approach. To add all program dependencies as exclude rules it seemed too complex and for that, we have sync of a program.**

- [x] ProgramStageSection 

**A program stage section has no sense without a program and program stage with data elements. Program and program stage sync creates data elements if does not exists on the receiver but this sync fail if not exists program, program stage or data elements on in the receiver**

**Web Api Bug: Works if you create new sections but if you exchange data elements between two program stage sections make strange behaviours**

- [x] User 

**if a user is created without org unit then the user won't be visible on users list because this dhis2 page use api/users endpoint with filter userOrgUnits=true (only play servers???) and only returns users with org unit.
User is created without password and an admin user have to assign a temporal password to the user to can make login with the user.**

#### Medium (6) 
- [x] DataEntryForm 

**This sync independently is only valid to edit DataEntryForm on the receiver because to add, the custom form is created in the receiver but It's not linked to data set because the web api for a DataEntryForm does not return related data set to include it as to include rule**.

- [ ]  ~~JobConfiguration~~ For the moment we don't add it

until now, add new metadata models is realized added new models in the metadata file, add a new entry in metadata array in the factory file and testing the metadata sync with this new model. The code is the same for all cases using current metadata sync logic using metadata endpoint to get and post. Retrieve all selected metadata using get. 
Job configurations are not retrieved from the metadata endpoint 

**Option 1**: we should create specific get request for every metadata sync to know if any metadataId is a JobConfiguration then make a request to the specific endpoint to get and post
**Option 2**: create a new section as delete objects for JobConfigurations we should create a new sync process type to existed (metadata, events, aggregated, -> jobConfigurations)

- [x] LegendSet

- [x] Section  

**A section has no sense without a data set with data elements. Data set sync creates data elements if does not exists on the receiver but this sync fail if not exists data set and data elements on in the receiver**

**Web Api Bug: Works if you create new sections but if you exchange data elements between two  sections make strange behaviours**

- [x] TrackedEntityAttribute
- [x] TrackedEntityType
- [x] UserRole -> In the dropdown appears as User Authority Group because is the display name

#### Simple (11)  
- [x] Category
- [x] CategoryCombo
- [x] CategoryOption
- [x] CategoryOptionCombo 
- [x] CategoryOptionGroup
- [x] CategoryOptionGroupSet
- [x] Option -> I have not added optionGroup as dependency rule because related optionGroup is not returned by web api
- [x] OptionGroup -> I have not added optionGroupSet as dependency rule because related optionGroupSet is not returned by web api
- [x] OptionSet
- [x] IndicatorType

### Pending refactors toward clean architecture
- Currently, in the d2.ts file, we are making transformations to metadata response from dhis2. 
  On the future when we have the code more refactored towards clean architecture, this transformations should be realized in the data layer in MetadataRepository implementation.
I have added a TODO comment in the code.

- Currently, metadata models defined in src/models/dhis/metadata.ts file are a kind of configuration for every metadata type where we configure things as:
   * include/exclude rules (domain configuration to know children dependencies of a metadata type to sync)
  * filters to show in table list by metadata type (presentation configuration)
  * columns to show in table list presentation (presentation configuration)
  * ...

We have to decide if all this configuration is domain o not. May has sense as domain configuration if we think that are settings of the app that potentially could be persisted in a data source and configurable by app or by the user but for the moment are always statically defined in the domain, but we need to talk it on future web meetings.
But if for us all it's domain configuration, this configuration should be in the domain layer without dhis2 and d2-ui-components dependencies. Currently exist dependencies with these two libraries in this configuration.

### :bookmark_tabs: Others

I have used in the PR d2-ui-componets version 1.3.1-beta.3
Alexis fixed a bug with data table columns in https://github.com/EyeSeeTea/d2-ui-components/pull/150


adrianq:
@xurxodev I am merging this now and I will review it in detail in the future. If I start testing this and asking for changes now it can be a nightmare to solve the conflicts later on

PR #547, opened 2020-07-09
Development who

adrianq:
This PR integrates the features developed for the WHO General maintenance section into the regular development branch. We are cutting off the three-headed hydra (development, development-malaria and development-who) and just keeping development.


SferaDev:
Not sure we want any of this on development

src/models/d2Model.ts
```diff
@@ -173,6 +173,7 @@ export class DataElementModel extends D2Model {
         "dataSets",
         "legendSets",
         "optionSets",
+        "optionSets.options",
```
SferaDev, line 176: Maybe this?

src/models/d2Model.ts
```diff
@@ -59,7 +59,7 @@ export abstract class D2Model {
         const filter = _.compact([
             search && isValidUid(search) ? `id:eq:${search}` : null,
             search && !isValidUid(search) ? `displayName:ilike:${search}` : null,
-            lastUpdatedDate ? `lastUpdated:ge:${lastUpdatedDate.toISOString()}` : null,
+            lastUpdatedDate ? `lastUpdated:ge:${lastUpdatedDate.format("YYYY-MM-DD")}` : null,
```
SferaDev, line 62: Maybe this?

adrianq:
Yes, let me finish the merge with development and we can evaluate if there is anything that worths to be merged

src/models/d2Model.ts
adrianq: Already in devel -> https://github.com/EyeSeeTea/metadata-synchronization/blob/development/src/models/dhis/metadata.ts#L81

src/models/d2Model.ts
adrianq: Not sure about this @SferaDev I guess the equivalent code is https://github.com/EyeSeeTea/metadata-synchronization/blob/development/src/models/deletedObjects.ts#L79 but not sure if there is anything we would like to migrate. Can you shed some light on this?

SferaDev: We don't have any ``toISOString()`` anymore, so it's safe to asume we already patched this (lastUpdated using time/date instead of only date)

adrianq: Great! Thanks!

adrianq:
No relevant work here so just closing it now that we have double checked there is no value in development-who

PR #554, opened 2020-07-14
[Feature] Implement changes to store page

SferaDev:
### :memo: Implementation

- Rename "personal access token"
- Remove mandatory check on store page
- Add check for empty token
- Add help button to header
- Add clear button next to test configuration

### :art: Screenshots

image


adrianq:
@SferaDev Just something very minor, after clicking 'Reset' maybe is better to keep the user in the same page instead of sending him back to the landing page, what do you think?

SferaDev:
> @SferaDev Just something very minor, after clicking 'Reset' maybe is better to keep the user in the same page instead of sending him back to the landing page, what do you think?

Sure thing, taking note and will implement on latest open PR.

PR #556, opened 2020-07-21
[Feature] Add pull request notifications

### LINKED ISSUE #534: Pull Request Metadata
### Original tasks

- [x] Set responsible user or user group 
   - [x] Add contextual action in the manual metadata sync or in a new section to add responsible
   - [x] Contextual action opens a dialog to select user and/or user group from the instance
   - [x] Save responsible linked to a piece of metadata (only Program and dataset)
- [x] On sync, check if there is a metadata or data responsible. If so create a Pull Request, otherwise it should work as now.
   - [x] Create Pull Request on remote instance. 
   - [x] Show form in a dialog with fields: description, author, date, set of metadata or data, other CC user or user group notifier
   - [x] Save Pull Request in data store both in origin and destination. Save JSON in remote for now (you are not allowed to get it).
   - [x] Send internal messaging/email to CC user or user group notifier
- [x] Notifications
   - [x] New section in landing page
   - [x] Show current notifications with the following actions:
      - [x] List notifications in a table with the following columns: read/unread, data/metadata, owner, timestamp, description, status (pending, approved, rejected), sent/received
      - [x] Filters: by status (dropdown), by read/unread, search box
      - [x] On init, check status of sent pull request asking to the remote.
         - [x] If approved, trigger sync
         - [x] If not, update pull request status
      - [x] Actions in remote:
         - [x] Show details. Summary listing the package
         - [x] Approve
            - [x] Update pull request status in data store
            - [x] Save results on history in local
         - [x] Reject
            - [x] Update pull request status in data store
         - [x] Delete (only available for the local)

### Feedback

- [x] Check same user permissions on different instances
- [x] Add pull then pull request dialog if same user is responsible
- [x] Check contextual actions and permissions
- [x] Check role CONFIGURATOR for responsibles add/remove and list in landing page
- [x] Send emails when importing pull request
- [x] Remove "delete" action
- [x] Add cancel in local
- [x] Ensure payload is cleared after successful import
- [x] After updating responsibles, update notifications PENDING (and create new prop on dataStore (responsibles)

SferaDev:
### :pushpin: References

-   **Issue:** Closes #534

### :memo: Implementation

- Add entities for PullRequestNotification (both Sent and Received)
- Add new notifications page with unread count on landing page
- Check if there're responsibles for a given metadata on SyncRule, ManualSync or Modules and show pull request creation dialog
- Add use cases and communication between instances and their notifications

### :art: Screenshots

image
image
image
image
image


adrianq:
@SferaDev Lot of new features, amazing work! I have noticed a few things though. I am dropping a few of them here but I will test the PR a bit more yet.

- [ ] Now that synchronization can happen in any direction maybe it makes sense to show in the summary the origin and the destination instances. Something like: 'Origin: http://dev.est.com:8085 -> Destination: http://dev.est.com:8086'

- [ ] What about adding a sort of help message in the PR dialog? A subtitle saying sth like 'There are responsibles for the selected metadata. You can still create a pull request and once approved by the responsible the synchronization will happen'

- [ ] Not sure about the internal message content:
image
Instead of 'user' I would say 'Created by'. At this point, can we differentiate between responsible and cc'ed users? What about linking at list the notification page instead of the application URL landing page?

- [ ] I think we discussed this but maybe you added it in another PR. What about showing in bold the notification row when it has not been read?

- [ ] When a PR has been accepted, the user needs to go into the notification and click on 'import'. This is fine but we may want to show somehow to the user that an action is missing. A sort of icon similar to the one when there is a conflict with a hover text?

- [ ] In the notifications, we are missing the destination or origin instance column, right? I mean the other instance that is involved in the synchronization process.

- [ ] I am getting an error when I try to import an accepted PR:
image
This is 8085 and I will keep it open so you can also review it:
image
Will try to test it with another set of metadata anyway.

SferaDev:
> - [ ] Now that synchronization can happen in any direction maybe it makes sense to show in the summary the origin and the destination instances. Something like: 'Origin: http://dev.est.com:8085 -> Destination: http://dev.est.com:8086'

Sure thing, I believe I added that as a comment somewhere. 

> - [ ] What about adding a sort of help message in the PR dialog? A subtitle saying sth like 'There are responsibles for the selected metadata. You can still create a pull request and once approved by the responsible the synchronization will happen'

Looks good to me, will do. 

> - [ ] Not sure about the internal message content:
> image
> Instead of 'user' I would say 'Created by'. At this point, can we differentiate between responsible and cc'ed users? What about linking at list the notification page instead of the application URL landing page?

Yes we should be able to differentiate from both of them and we can add it to the message. 

The linking requires to add a query param but it's feasible to do it. 

> - [ ] I think we discussed this but maybe you added it in another PR. What about showing in bold the notification row when it has not been read?

We did discuss that but I completely forgot about that point. 

> - [ ] When a PR has been accepted, the user needs to go into the notification and click on 'import'. This is fine but we may want to show somehow to the user that an action is missing. A sort of icon similar to the one when there is a conflict with a hover text?

Sure thing. 

> - [ ] In the notifications, we are missing the destination or origin instance column, right? I mean the other instance that is involved in the synchronization process.

Will add the column and details. 

> - [ ] I am getting an error when I try to import an accepted PR:
> image
> This is 8085 and I will keep it open so you can also review it:
> image

Strange, this looks like a problem generating the payload for that dataSet. Will take a look even though apparently unrelated to the notifications feature.

PR #575, opened 2020-08-14
[Fix] Workaround 414-uri too long

### LINKED ISSUE #250: Selection in other pages produces 414 URI too long
https://github.com/EyeSeeTea/metadata-synchronization-blessed/issues/66

We have this problem with all the ``id:in:[]`` filters (selectedItems and filterRows).

### LINKED ISSUE #532: Package comparison
Packages -> Remote instance -> Contextual Action: Package comparison: MetadataComparison.tsx

`MetadataComparison.diff(localPackageNotSaved, remotePackage) -> {updated: ..., created: ...}`

- localPackage using IDS from remotePackage.

Front-end. Modal rendering a summary of:
   - [x] Green: added objects
   - [x] Orange: modified objects

Layout:
- [x] New contextual action on package view 'Compare with local instance'
- [x] Open a modal dialog fullscreen
- [x] For now show only difference (no need for two sections). For example
   - 2 New DEs (include name and some relevant fields)
   - 3 Deleted DEs (include name and some relevant fields) -> NOT POSSIBLE
  - 4 Modified Dataset (include name and some relevant fields)
- [x] Two buttons at the bottom: Import / Cancel
- [x] Check if import is already done and we can reuse the method -> ./domain/metadata/usecases/ImportMetadataUseCase.ts
- [ ] Test for getMetadataDiff -> NOT FOR NOW.
- [x] Check performance with orgUnits

Later:
Left-side: remote instance → Right-side: local instance

tokland:
### :pushpin: References

-   **Issue:** Closes #250 

### :memo: Implementation

- On `MetadataD2ApiRepository` there were 2 methods affected: `listMetadata` and `listAllMetadata`. I've added two new wrapper methods (`getListAll`, `getListPaginated`) that check the length of the filter `id->in:[ids]` and, if greater than a max value (500): 1) split the list, 2) perform N requests, 3) join the objects, 4) sort the objects (in both cases, api/manual), 4) paginate them (if needed)

- While testing, I noticed that the current page was not reset when filters changed. I don't know if that was the indented behaviour, but I found it problematic, at least for the "only selected items" so I force a `page: 1` for that filter. @SferaDev, can you confirm? Do we need this for the other filters? https://github.com/EyeSeeTea/metadata-synchronization/pull/575/files#diff-e52fd5e6cefafb73b3ea3210d73e7149R187

### DIFF
--- a/src/data/metadata/MetadataD2ApiRepository.ts
+++ b/src/data/metadata/MetadataD2ApiRepository.ts
@@ -309,7 +309,7 @@ const defaultOrder = { field: "id", order: "asc" } as const;
 interface GetListAllOptions {
     type: ListMetadataParams["type"];
     fields: object;
-    filter: Dictionary<FilterValueBase>,
+    filter: Dictionary<FilterValueBase>;
     order?: ListMetadataParams["order"];
 }
 
--- a/src/data/metadata/MetadataD2ApiRepository.ts
+++ b/src/data/metadata/MetadataD2ApiRepository.ts
@@ -1,3 +1,4 @@
+import { FilterValueBase } from "d2-api/api/common";
 import _ from "lodash";
 import moment from "moment";
 import { Ref } from "../../domain/common/entities/Ref";
@@ -20,12 +21,12 @@ import { TransformationRepository } from "../../domain/transformations/repositor
 import { D2Api, D2Model, MetadataResponse, Model, Stats } from "../../types/d2-api";
 import { Dictionary } from "../../types/utils";
 import { cache } from "../../utils/cache";
+import { promiseMap } from "../../utils/common";
+import { paginate } from "../../utils/pagination";
 import {
     metadataTransformationsFromDhis2,
     metadataTransformationsToDhis2,
 } from "../transformations/PackageTransformations";
-import { promiseMap } from "../../utils/common";
-import { paginate } from "../../utils/pagination";
 
 export class MetadataD2ApiRepository implements MetadataRepository {
     private api: D2Api;
@@ -162,14 +163,14 @@ export class MetadataD2ApiRepository implements MetadataRepository {
         filterRows,
         search,
     }: Partial<ListMetadataParams>) {
-        const filter: Dictionary<unknown> = {};
+        const filter: Dictionary<FilterValueBase> = {};
 
         if (lastUpdated) filter["lastUpdated"] = { ge: moment(lastUpdated).format("YYYY-MM-DD") };
         if (group) filter[`${group.type}.id`] = { eq: group.value };
         if (level) filter["level"] = { eq: level };
         if (parents) filter["parent.id"] = { in: cleanOrgUnitPaths(parents) };
-        if (showOnlySelected) filter["id"] = { in: selectedIds };
-        if (filterRows) filter["id"] = { in: filterRows };
+        if (showOnlySelected) filter["id"] = { in: selectedIds.concat(filter["id"]?.in ?? []) };
+        if (filterRows) filter["id"] = { in: filterRows.concat(filter["id"]?.in ?? []) };
         if (search) filter[search.field] = { [search.operator]: search.value };
 
         return filter;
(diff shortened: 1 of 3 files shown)


EyeSeeTeaBotTest:
<h3>Tests failed with errors :rotating_light:</h3><br />Build commit Merge beb4dcb105bbe0d10dba4e2e81176fb49fcb2e5c into 6b249e1dfc599f3b4822cd5b3712567dab707b2c with commit id 84fb7412740b342b108f514348cb9bad35d9e45e on branch feature/workaround-uri-too-long-250<br /><br />Travis log available at <a href=https://travis-ci.org/EyeSeeTea/metadata-synchronization/builds/717885677 target=_blank>builds/717885677</a><br /><br />See cypress results at <a href=https://eyeseetea.github.io/feedback-ci/EyeSeeTea/metadata-synchronization/pr/575/84fb7412740b342b108f514348cb9bad35d9e45e target=_blank>feedback-ci/EyeSeeTea/metadata-synchronization/pr/575/84fb7412740b342b108f514348cb9bad35d9e45e</a>

src/data/metadata/MetadataD2ApiRepository.ts
```diff
@@ -247,3 +300,43 @@ const formatStats = (stats: Stats) => ({
...
+
+// (max_size=8000 - rest_url=1000) / (id_length=11 + comma_encoded=3)
+const maxIds = 500;
+
+const defaultOrder = { field: "id", order: "asc" } as const;
+
+interface GetListAllOptions {
+    type: ListMetadataParams["type"];
+    fields: object;
+    filter: Record<string, unknown>;
```
SferaDev:
```suggestion
    filter: Dictionary<FilterValueBase>,
```

Not sure why ``Dictionary<T>`` which is a ``Record<string, T>`` works but ``FilterBase`` does not.

src/data/metadata/MetadataD2ApiRepository.ts
```diff
@@ -247,3 +300,43 @@ const formatStats = (stats: Stats) => ({
...
+
+type GetListGenericResponse =
+    | { useSingleApiRequest: false; objects: unknown[] }
+    | { useSingleApiRequest: true; order: string };
+
+function getIdFilter(
+    filter: Record<string, unknown>,
+    maxIds: number
+): { inIds: string[]; value: object } | null {
+    const inIds = filter && filter["id"] ? (filter["id"] as { in?: string[] })["in"] : undefined;
```
SferaDev:
```suggestion
function getIdFilter(
    filter: Dictionary<FilterValueBase>,
    maxIds: number
): { inIds: string[]; value: object } | null {
    const inIds = filter ? filter["id"] : undefined;
```

With a more significant type we can simplify the expression a lot

tokland: Cool. I was unsure why we were using `unknown`, but I didn't want to tweak with the types, not it looks better. I've also simplified the conditional, now with better types is much more simple: b76efd19

src/data/metadata/MetadataD2ApiRepository.ts
tokland: Do you mean `Record<string, FilterValueBase>`? Weird, should work.

SferaDev:
My bad I forgot some ``console.log`` in the code I pushed.

Everything else LGTM.

About resetting page to 1 on filters change, it makes sense and it's probably a "bug" we have from long time. We reset the page for some but not all of them right now. I would however leave the decision to @adrianq to see if we leave as-is or review all the filters and homogenize their behavior.

src/data/metadata/MetadataD2ApiRepository.ts
SferaDev: No ``Record<string, FilterValueBase>`` works too, I tend to use ``Dictionary`` however as it's more verbose about what it returns.

I meant that using ``FilterBase`` does not work (errors in the ``in`` access)

tokland: Ah, now I understand what you mean. The problem is the full filter is:

```
export interface FilterBase {
    [property: string]: FilterValueBase | FilterValueBase[] | undefined;
}
```
So can have either `filter["id"].in` or `filter["id"][0].in` depending whether is an array or not. Now we are considering only the non-array case.

adrianq:
@tokland @SferaDev I have created #584 so we review the filters' behaviour in all pages. I agree we should reset page to page 1 when filters change.

PR #608, opened 2020-08-27
Refactor from/to instance selectors

### LINKED ISSUE #602: Make origin and destination instance more clear
In the Sync sections:
- [x] As the users enters each section, he or she should be asked from WHERE to WHERE he/she wants to synchronize data. If the users select another instance as FROM the TO is automatically This instance. If the users selects from THIS instances he/she can select anything for TO:. 
 
Instance mapping:
- [x] Replace text on top of the mapping window to say: Between this instance and XXXXXX”
- [x] Same on each of 4 children pages

---
Later: Multiselector for TO

### LINKED ISSUE #620: Catch and show error 409
- [x] Detect errors during payload building (ie. 409 fetching dataValues)
- [x] Show specific error when a 409 happens on import (Example program not assigned to an org unit)

tokland:
### :pushpin: References

-   **Issue:** Closes #602

### :memo: Implementation

There are some details that can be implemented differently, some notes about them:

- No labels for from/to, it didn't look good. Let's see if the arrow is clear enough.
- Auto-select first instance in the destination selector (on the first render and on source instance change)
- Never hide the step "instance selection", as even for local instance there are advanced import options that the user may want to change. In this case, it shows: `Destination: This instance`.
- Not done for now: destination selector should ideally be multiple-selection, but we need a different component for this (Autocomplete) which should be styled to look similar. [let me know if this is a must].
- A change on remote instance on the step is reflected in the page selector.

### :art: Screenshots

image
image


SferaDev (approved):

adrianq:
No worries about the multiple-selection dropdown. It was a nice to have but it is not particularly relevant. I have made some cosmetics changes but I am not particularly happy with them. Will give a shout to Ray and ask him for some magic!

PR #611, opened 2020-08-28
[Feature] Update permissions modules

### LINKED ISSUE #561: Roles/permissions for modules/packages
- [x] Executors can also see the package table
- [x] Executors can only download to json in package table.
- [x] Add sharing settings to modules so that configurators only see their modules.
  - [x] Add components to ModulesListTable
  - [x] Update model to validate permissions before executing actions (write)
  - [x] Update list use case to show only visible modules
  - [x] Update list use case to show only visible pacakages
- [x] Disabled editing department on module (after saving)

Details:
  - [x] Packages inherit sharing settings from their modules
  - [x] Sharing settings by default are public (rw) and department (rw)

SferaDev:
### :pushpin: References

-   **Issue:** Closes #561 

### :memo: Implementation

- Disable department change on module edit
- Show package listing to app executors on landing page
- Add sharing settings for modules and respect package listing depending on its module

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2020-08-28T07:25:40.689Z\n"
-"PO-Revision-Date: 2020-08-28T07:25:40.689Z\n"
+"POT-Creation-Date: 2020-08-31T07:36:10.258Z\n"
+"PO-Revision-Date: 2020-08-31T07:36:10.258Z\n"
 
 msgid "Field {{field}} cannot be blank"
 msgstr ""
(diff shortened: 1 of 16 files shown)


adrianq:
@SferaDev This looks good. Just a couple of things:
- Can we show an asterisk in the Department to make clear the field is compulsory?
- The department is not added by default as r/w. I am using a non-super admin user in case it helps...
- Question: We are not storing the sharing settings for the packages in the data store, right? We read the sharing settings from the modules on run time, correct?

SferaDev:
> @SferaDev This looks good. Just a couple of things:
> - Can we show an asterisk in the Department to make clear the field is compulsory?

Sure thing

> - The department is not added by default as r/w. I am using a non-super admin user in case it helps...

Actually the department is not a user setting that can be added or removed but a check like the admin check. When you said is not added I guess you mean that it doesn't appear in the sharing setting dialog but the rights still apply to any department's users? We can add it to the dialog and show it as non removable.

> - Question: We are not storing the sharing settings for the packages in the data store, right? We read the sharing settings from the modules on run time, correct?

Exactly

PR #615, opened 2020-09-04
[Fix] Add localhost validation to field

SferaDev:
### :pushpin: References

-   **Issue:** Closes https://github.com/EyeSeeTea/metadata-synchronization-blessed/issues/73

### DIFF
--- a/src/domain/common/entities/Validations.ts
+++ b/src/domain/common/entities/Validations.ts
@@ -15,7 +15,7 @@ export interface ModelValidation {
 }
 
 // Diego Perini (License: MIT)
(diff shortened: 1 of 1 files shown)


adrianq:
@SferaDev  This works if the URL is localhost:8080 but still failing if the URL is http://localhost:8080 which is btw the one used by the client and reported as an issue

SferaDev:
> @SferaDev  This works if the URL is localhost:8080 but still failing if the URL is http://localhost:8080 which is btw the one used by the client and reported as an issue

No problem we can add an optional capture group with http(s)

PR #617, opened 2020-09-04
Metadata filter rules

### LINKED ISSUE #545: Add new pattern metadata selection for sync rules
- [x] From branch feature/update-permissions-modules
- [x] New step 3 in metadata sync rule wizard: "Filter rules"
  - [x] Current metadata step won't be mandatory, however at least one of the metadata selection should happen (validate on step 3->4 transition)
  - [x] Object table showing available filters, one filter per row.
  - [x] Button to create new filter at the top-left corner. It opens a modal containing filter component
  - [x] Delete
  - [x] Edit
  - [x] Available filters:
    - [x] Created -> Absolute Date / Relative Date (Use Period - step 4 in aggregate). Substitute 'All periods' by 'all time'
    - [x] Last Updated -> Absolute Date / Relative Date (Use Period - step 4 in aggregate). Substitute 'All periods' by 'all time'
    - [x] Regex-like filter for name/code/*description -> Textfield with DHIS2 format (begins with, ends with, contains)
    - [x] Metadata Type -> Dropdown
    - [] Later:
    - [] Created by user or group -> Sharing Setting Component 
    - [] Last Updated by user or group -> Sharing Setting Component 
    - [] Sharing settings assigned to user or user group -> Sharing Setting Component
  - [] Later - Show results applying filter in the pop-up inside an object table

- [x] Summary: show filters (without results), keep specific metadata shown from step 2
- [x] Save filters to dataStore
- [x] Model: final metadata is metadata step 2 UNION result from step 3

tokland:
### :pushpin: References

-   **Issue:** Closes #545

### :memo: Implementation

- Sync: to apply the string filter on code/name/description we need 3 separate requests for each filter rule, as we are doing an AND for all the other conditions but an OR for those. If there is no string filter, perform only one request. Note that if we didn't care about "description", we could use virtual field "identifiable" (id+name+code)
- @SferaDev, I am reusing `buildPeriodFromParams` I don't really need those default start/end dates, but I was 
weary of re-implementing that logic. Let me know if you'd take another approach.
- New entity: `FilterRule`. I've taken the interfaces + free functions approach (for no particular reason).
- New hook: `hooks/useOpenState.ts`. It tries to capture the logic around having a modal UI associated with data (example: open a dialog to create/edit an entity)
- Abstracted `PeriodSelection.tsx` (used in `sync-wizard/data/PeriodSelectionStep.tsx` and now on the date fields of a filter rule) -> **Testers should check that this step still works normally: Edit aggregated synchronization rule -> Periods**
- Validations appear in snackbar messages.
- New generic component: `TextFieldOnBlur.tsx`: A normal TextField notifies changes to the parent on every keystroke. This is usually slow and unnecessary. Instead, **this component notifies the parent only on the blur event** of the component. It requires a somewhat tricky logic to both keep an internal state and react to parent props.value changes, but it seems to work fine. 
- @SferaDev: Check new generic type helpers in `src/types/utils.ts`, they may come handy. `buildObject` is just a wrapper trick so TS can infer the key values on a `Record` while only passing the type value.
- Use https://github.com/EyeSeeTea/feedback-tool/pull/7
- Cypress specs: I have problems with data-test matches. Though untested, I've modified the specs anyway. @SferaDev 

### :art: Screenshots

Screenshot from 2020-09-04 10-42-40

Screenshot from 2020-09-04 10-42-57

Screenshot from 2020-09-04 10-43-13


adrianq:
@tokland When I try to create a metadata rule I am getting this error all the time:
image

PR #619, opened 2020-09-07
[Fix] Bugs found after release

SferaDev:
### :memo: Implementation

- Add back ``dataSetElements`` to ``dataSet`` during metadata sync
- Do not hang-up if user does not have permissions to retrieve dataValues
- Fix cosmetic bug editing name or description of a sync rule

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "metadata-synchronization",
     "description": "Advanced metadata & data synchronization utility",
-    "version": "2.1.0",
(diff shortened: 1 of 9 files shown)


SferaDev:
Thanks @idelcano for noticing this quickly, @adrianq I believe we should issue a quick patch release to fix this.

The reason is because on ``d2`` the ``dataSetElements`` was a metadata model and had a schema so it wasn't needed to translate the property to a valid schema.

Issue #624, opened 2020-09-10
Review how to select root orgUnit in mapping


SferaDev:
- [x] For the selected orgUnit, the table should render also this parent orgUnit as the first row (as maintenance-app does)
- [x] When filtering, parent row should be also filtered.

PR #638, opened 2020-09-18
Organisation units table fixes

### LINKED ISSUE #637: Review search function for Org Unit
- [x] Fix error on orgUnit search (`Cannot read property 'some' of undefined`)
- [x] When no orgUnit is select, a filter `parent.id:in:[]`is applied, shoud be no filter to show results.

See https://github.com/EyeSeeTea/metadata-synchronization-blessed/issues/74

Although not strictly related, review  this one as well https://github.com/EyeSeeTea/metadata-synchronization-blessed/issues/75

tokland:
### :pushpin: References

-   **Issue:** Closes #637, #624 
- Requires d2-ui-components: https://github.com/EyeSeeTea/d2-ui-components/pull/179 (already as beta)

### :memo: Implementation

- Don't pass empty refs to the `parents.id` filter. 
- Added a generic type guard helper `isNotEmpty` (`isNotEmpty<T>(xs: T[] | undefined): xs is NonEmptyArray<T>`).
- Add parent orgUnit on metadata table as a virtual row, as maintenance-app does, pagination is not affect. That applies for every orgUnit metadata table component. **Or do we need only for mapping?**

image


tokland:
Ready for review. @SferaDev there is a bit of tricky logic in 24ea68c8 with the filters, but should work well. I am using that addition we made to d2-api so filters can be arrays.

src/data/metadata/MetadataD2ApiRepository.ts
```diff
@@ -64,18 +64,14 @@ export class MetadataD2ApiRepository implements MetadataRepository {
...
-        order,
-        ...params
-    }: ListMetadataParams): Promise<ListMetadataResponse> {
+    public async listMetadata(listParams: ListMetadataParams): Promise<ListMetadataResponse> {
+        const { type, fields = { $owner: true }, page, pageSize, order, ...params } = listParams;
         const filter = this.buildListFilters(params);
         const { apiVersion } = this.instance;
         const options = { type, fields, filter, order, page, pageSize };
-        const { objects, pager } = await this.getListPaginated(options);
+        const { objects: baseObjects, pager } = await this.getListPaginated(options);
```
SferaDev, line 72: Pager does not include `parentObjects`

src/presentation/react/components/metadata-table/MetadataTable.tsx
```diff
@@ -405,10 +405,11 @@ const MetadataTable: React.FC<MetadataTableProps> = ({
 
     useEffect(() => {
         if (model.getCollectionName() === "organisationUnits" && !filters.parents) return;
+        const fields = model.getFields();
 
         setLoading(true);
         compositionRoot.metadata
-            .list({ ...filters, filterRows, fields: model.getFields() }, remoteInstance)
+            .list({ ...filters, filterRows, fields, includeParents: true }, remoteInstance)
```
SferaDev, line 412: Shouldn't it be includeParents true if model is organisationUnits and false otherwise?

SferaDev (commented):
Couple of comments

src/data/metadata/MetadataD2ApiRepository.ts
tokland: Yes, that was intended, it simplifies things and maintenance-app does the same. @adrianq, do we add it?

image

src/presentation/react/components/metadata-table/MetadataTable.tsx
tokland: I thought the idea was to keep it as generic as possible, in case in the future any model has parents. Anyway, it should be ignored for any other model since `parents` will be undefined, right?

SferaDev: If we follow that logic the `includeParents` is redundant. It could be removed completely and we would still fallback to the isNotEmpty check.

In any case, @adrianq this is not very important and the PR can be merged anyway.

tokland: I see your point, I was thinking of another potentiall call to the use case without that option, but probably use cases do not get to reused like this.

src/data/metadata/MetadataD2ApiRepository.ts
adrianq: No, let's keep it as is for now

src/presentation/react/components/metadata-table/MetadataTable.tsx
adrianq: As discussed, as we don't have a strong opinion I would leave it for the next one changing this code.

PR #695, opened 2020-11-12
[Refactor] Add new store repository

### LINKED ISSUE #694: Refactor store domain to uncouple use cases from the data store of Dhis2
- [x] Rename StorageRepository to DataStore or similar
- [x] Move Namespaces of data store to data layer
- [x] Create StoreRepository interface (list, get, save, delete, getDefault)
- [x] Create StoreDataStoreRepository (list, get, save, delete, getDefault)
- [x] Use store repository from GetStoreUseCase
- [x] Use store repository from ListStoresUseCase
- [x] Use store repository from DiffPackageUseCase
- [x] Use store repository from DownloadPackageUseCase
- [x] Use store repository from GetStorePackageUseCase
- [x] Use store repository from ListStorePackagesUseCase
- [x] Use store repository from PublishStorePackageUseCase
- [x] Use store repository from SaveStoreUseCase

SferaDev:
### :pushpin: References

-   **Issue:** Closes #694

### :memo: Implementation

- Add new repository and update use-cases
- Refactor old usages to follow a more clean approach

- No behavior changes should be found

### :bookmark_tabs: Others

Since this is a big PR, it probably makes sense to skip ``2.5.0``, be merged with the changes of the rest of feedback and be included for next release. I will base off my rest of changes on top of this though to minimize the conflicts on my side.


SferaDev:
All changes are in, I still need to do a full test of this refactor but I will do it while developing the next feature.

SferaDev:
Noticed two more use cases that can be included in this refactor, moving back to draft until later today.

xurxodev (changes requested):
Thanks, @SferaDev, Seems a very big pull request I suggest you a better approach to realizing smaller pull requests in the future to more secure tests by @adrianq to merge.
After a minor review, I detect two bigger features:
* **Uncouple data store from use cases related to store domain entity**
This was the principal feature of the issue and seems the code is ok to our talk in the last pair programming meeting but it's difficult to be sure with so much code changed.
* **Avoid duplicate code in all uses cases to retrieve repositories**.
You have used the inheritance to solve this problem with duplicate code but normally is better composition over inheritance, because it's more versatile and the inheritance couple the use cases with repositories that it may be not needed and we have potential problems in the future:
  * for example creating a test for unit tests in the future when we start to create this type of test. 
  * Maybe exist use cases that receive the repository directly in the constructor and this use case base is not necessary or only partially, this is confusing.

For me has more sense the code in DefaultUseCase is in the RepositoryFactory because its behavior is more like a factory creating instances and cache it. In the future, we can split into different factories if is very big using composition but you cannot inherit from more the one base class by default. 

Anyway, we will take in the next pair programming meeting.

SferaDev:
> Thanks, @SferaDev, Seems a very big pull request I suggest you a better approach to realizing smaller pull requests in the future to more secure tests by @adrianq to merge.

Yes, it seems a pretty big PR but if you take a look more than a half of the changes are part of the renames on either files, folders and their relative imports (affecting over 55 of the 93 changed files). With an app of these dimensions doing a refactor that renames essential pieces without touching dozens of files is very unlikely.

> * **Uncouple data store from use cases related to store domain entity**
>   This was the principal feature of the issue and seems the code is ok to our talk in the last pair programming meeting but it's difficult to be sure with so much code changed.

Yes, everything is in there, and you can see it in the diff for the ``packages``, ``store`` and ``data`` folders.

> * **Avoid duplicate code in all uses cases to retrieve repositories**.
>   You have used the inheritance to solve this problem with duplicate code but normally is better composition over inheritance, because it's more versatile and the inheritance couple the use cases with repositories that it may be not needed and we have potential problems in the future:

Done, moved from ``DefaultUseCase`` to ``RepositoryFactory``

PR #732, opened 2020-11-24
Feature/add documents sync

### LINKED ISSUE #586: Add file resources sync
- [x] Read origin metadata from document
- [x] Read origin ``/api/documents/id/data`` file
- [x] Create fileResource on destination ``curl -X POST -F "file=@file;filename=file"`` with file and read fileResourceId (https://docs.dhis2.org/2.30/en/developer/html/webapi_file_resources.html)
- [x] Create/update document with metadata from origin (same id) and update url to the newly created fileResourceId

Error handling
- [x] Document data does not exist on origin
- [x] Document too big
- [x] Rest of errors related to import

xurxodev:
### :pushpin: References

-   **Issue:** Closes #586 

### :memo: Implementation

To sync documents, we can use metadata endpoint and It's necessary to create an infrastructure to create blob data of documents in target previously to sync.

- [x] create DocumentsModel and to add it in metadata models
- [x] create File repository to read from `/api/documents/id/data`and save in `/api/fileResources`
- [x] Modify metadata sync use case to create blob data of documents in target instance previously and update metadata package documents to assign the new resource file id 

I have added the mime-type lib to get the extension of a mime type.

I have created the FileRepository using fetch because d2api has not this feature to read and write blob data.

I have implemented the file repository to download and upload the resource files using fetch because d2api has not this feature. I have created an issue in d2 api https://github.com/EyeSeeTea/d2-api/issues/80

### :bookmark_tabs: Others


adrianq:
@xurxodev  PR seems to be working fine both with doc and external links. Just something minor, 'Documents' doesn't seem to be a very intuitive name, right? Maybe we can use 'Files' instead. I am merging this anyway. We can tackle the PR comments once I have gone over all the PRs

PR #733, opened 2020-11-24
Feature/add contextual action create module from package list

### LINKED ISSUE #709: Create Module from package

xurxodev:
### :pushpin: References

-   **Issue:** Closes #709 

### :memo: Implementation

- Add contextual menu generate module from the package list

This action only is visible from the package list page when the package is created from a file and status is equal to install.

This action set the autogenerated property to false of the module created during the create package from file action.

Currently, the modules autogenerated from the file contains all level 1 metadata ids as selected. When this module is marked as non-autogenerated using the generate module from the package, it's possible that appears metadata ids as selected and these items are not visible in the metadata step in the module because the module type does not exist in the dropdown.

For example:
Creating a package from a file with a SQL view and then generate a module from the package
Captura de pantalla 2020-11-24 a las 10 49 42

### :bookmark_tabs: Others


adrianq:
I will test this together with https://github.com/EyeSeeTea/metadata-synchronization/pull/737 as it doesn't make too much sense to validate it alone. Something we already discussed and we may need to solve in the short term:
- [ ] Filter level-1 ids by the models that can be part of a package, i.e. Dashboard, Program and Dataset

PR #736, opened 2020-11-25
[Feature] Allow building multiple front-ends

SferaDev:
### :memo: Implementation

- We need to build multiple front-ends for both malaria and MSF
- Extend the widget generator to include multiple applications
- This is a basic approach that can be improved in the future

Now we only have two customizable commands:

- ``yarn start [variant] -p 8081`` that starts the dev server (if no variant sent main app is ran)
- ``yarn build [variant]`` that builds the release files (if no variant sent all front-ends are built)


adrianq:
Merging this. I will validate this PR together with https://github.com/EyeSeeTea/metadata-synchronization/pull/738

PR #737, opened 2020-11-26
Feature/add save and import to import from file dialog

### LINKED ISSUE #697: Add new button in dialog to create a package from a file
Add new button: Save & import. It should simply open the import wizard once the package has been generated

xurxodev:
### :pushpin: References

-   **Issue:** Closes #697 

### :memo: Implementation

- Add Save & Import button to create a package from a file dialog. I have used material-ui dialog instead of confirmation dialog of d2-ui-components because it's a very specific use case with two accepts buttons that confirmation dialog does not support.
- I have fixed a bug where the package id created in the entity was overwritten in the save package use case.
- Show package import dialog after save & import. In this case, I have hidden the steps related to package selection (1 and 2).
- I have fixed a bug with step 3 info messages when the package does not contain any element to map, for example, if the package is created from a file and contains only a SQL view, step 3 shows now the message 'There are not elements to map in the package'.
- I have created a new metadata owner type named package that it's not saved in the data store. It's saved temporally in memory in the package import rule entity and used during the import process.

### :art: Screenshots

Captura de pantalla 2020-11-26 a las 8 41 58
Captura de pantalla 2020-11-26 a las 8 42 29

### :bookmark_tabs: Others


adrianq:
Works like a charm @xurxodev

PR #790, opened 2020-12-18
[Fix] This instance not found on init

SferaDev:
### :memo: Implementation

- Add start-up routine on init


SferaDev:
@xurxodev Infrastructure question. This is a start-up routine (and we might have more), not sure about the best place to locate it. Adding it directly to ``CompositionRoot`` temporarily to fix the issue but asking for advice on where we should locate this kind of methods.

Note: In the future the permissions would be initialized similarly (roles and routes), they're not because we don't have them in clean architecture.

xurxodev:
@SferaDev composition root only should contain related code to dependencies.
App initializations are use cases. Currently, we are executing use cases from components because we are not using a presentation pattern like MVP, MVVM, MVI or BLoC then for coherence we should execute initializations use cases from the WebApp component because it's the start point component of the app.

xurxodev (approved):
Code has a sense for me. Just some a minor comment about naming, compositionRoot.presentation.initialize is confused because is not making anything about the presentation layer, I think a name like compositionRoot.app.initialize, compositionRoot.bootstrap.initialize or similar is more suitable.

PR #810, opened 2020-12-28
[Feature] Use user org units in MSF front-end

### LINKED ISSUE #756: Button to trigger sync rules will trigger a set of sync rules. (B10)
The set of sync rules to be synced will be based on an internal metadata rule using the OUs of the user.
- [x] Run sync rules that have org units where the user can access (Data Capture).
- [x] Check if analytics and any other relevant API call take into account the security rules
   - [x] If so, with at least one org unit accessible we will run the sync rule
   - [x] If not, remove org units not accessible by the user

SferaDev:
### :pushpin: References

-   **Issue:** Closes #756

### :memo: Implementation

- Execute rules relative to user org units


adrianq:
All good, just two minor things we can discuss tomorrow:
- We are finally using data view org unit which makes sense but it is not what we said in the issue description
- We may want to add a previous trace in the logging dialog saying something like 'Retrieving information from the system...'

PR #811, opened 2021-01-11
[Feature] Add minimum date for org units

SferaDev:
### :pushpin: References

-  **Issue:** CU-c8umqb

### :memo: Implementation

- Add new UI to MSFSettings
- Store minimum dates on dataStore
- Filter rules by period

- The UI does not list other minimum dates to the side


adrianq:
PR seems to be working fine: sync rules are split properly when needed and there is no problem with org unit visible by the user. The only thing missing here is to show a warning when an intersection is detected (last point in the ticket).

PR #836, opened 2021-03-03
Review security in MDsync

xurxodev:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/eh5vmr

### :memo: Implementation

- [x] Add namespace prop to manifest in the app variants 
- [x] Create migration to move username and password to details
- [x]  Remove instances- key in the migration
- [x] Show a message to the user during the migration about share settings for instances
- [x] Create or edit a new instance from the app should save user and password in the instance details key
- [x] Add level to migration messages and to add style to text in the component
- [x] Add a message in the migration screen of the app from 2.28 to 2.31 to warn the users of this potential risk
- [x] Add sharing settings per instances (contextual action)
- [x] Show a warning on click in share settings contextual action from instance list page if the dhis2 version is between 2.28 and 2.31 
- [x] Limit actions or catch errors by instance share settings
    - [x] All instance actions require write access but test connection that required read access
    - [x] Catch error to execute manual sync if you haven't read permissions about some target instance
    - [x] Catch error to execute sync rule if you haven't read permissions about some target instance
    - [x] Catch error to download json from a sync rule (summary step) if you haven't read permissions about some target instance
    - [x] Catch error to download json from sync rule list page if you haven't read permissions about some target instance
    - [x] In the instances dropdown shows only instances with read permissions for the user. This task includes refactor to move getCurrentUser method in instance repository to own UserRepository and returned all necessary for permissions (isAdmin, IsAppConfigurator, isAppExdecutor)

### :art: Screenshots

Show a message to the user during the migration about share settings for instances is realized according to the current architecture in migrations. Now messages send from migrations to component has a level prop and according to this level the message is rendered with a different style in the component

Migrations

- After adding the namespace in web assembly, I have tested several app installations from several variants. I was concerned about a text in the [documentation](https://docs.dhis2.org/en/develop/loading-apps.html#apps_creating_apps) about the namespace key because we have several apps using the same datastore namespace, but I have realized modifications in sync rules or modules in different installed variants and work successfully.

**Documentation text**: The namespace property can be added if your app is utilizing the dataStore or userDataStore api. When adding the namespace property, only users with access to your app are allowed to make changes to the namespace. A namespace can only be reserved in this way once. **If another app tries to reserve a namespace already in use, the installation of the other app will fail**.
**If you have a collection of apps that want to share the same namespace, but also wish to reserve it, the users of the apps needs to have the authority to use the app that initially reserved the namespace**.

### :bookmark_tabs: Others


EyeSeeTeaBotTest:
<h3>Tests failed with errors :rotating_light:</h3><br />Build commit Merge 0883088b420c1226bea42d7f87164a5396f089fe into 64b8ea0f8ff7c18afbfee60fc0e1051f4f181504 with commit id e596910c491e34f4ec030c4c2b3df19e9e452562 on branch feature/security<br /><br />Travis log available at <a href=https://travis-ci.org/EyeSeeTea/metadata-synchronization/builds/761247064 target=_blank>builds/761247064</a><br /><br />See cypress results at <a href=https://eyeseetea.github.io/feedback-ci/EyeSeeTea/metadata-synchronization/pr/836/e596910c491e34f4ec030c4c2b3df19e9e452562 target=_blank>feedback-ci/EyeSeeTea/metadata-synchronization/pr/836/e596910c491e34f4ec030c4c2b3df19e9e452562</a>

src/data/storage/StorageDataStoreClient.ts
```diff
@@ -72,4 +72,49 @@ export class StorageDataStoreClient extends StorageClient {
...
+            metadata,
+            "created",
+            "lastUpdated",
+            "lastUpdatedBy",
+            "namespace",
+            "key",
+            "value",
+            "favorite",
+            "id"
+        );
```
SferaDev, line 89: Not 100% in with the ``omit`` approach, it _should_ be safe to use if we are certain about the response of the API endpoint but it's normally safer to do the negate ``pick`` logic.

In the constant storage I've implemented it this way:

```ts
    public async getObjectSharing(key: string): Promise<ObjectSharing | undefined> {
        const {
            user,
            userAccesses,
            userGroupAccesses,
            publicAccess,
            externalAccess,
        } = await this.getConstant(key);

        return { user, userAccesses, userGroupAccesses, publicAccess, externalAccess };
    }
```

src/data/storage/StorageDataStoreClient.ts
```diff
@@ -72,4 +72,49 @@ export class StorageDataStoreClient extends StorageClient {
...
+            object,
+        }).response;
+    }
+
+    private async getMetadataByKey(key: string) {
+        const response = await this.api.get<MetadataDataStoreKey>(
+            `/dataStore/${dataStoreNamespace}/${key}/metaData`
+        ).response;
+
+        return response.data;
```
SferaDev, line 107: Instead of using ``response.data`` we have the helper method ``getData`` on the API:

```ts
        const data = await this.api
            .get<DataStoreMetadata>(`/dataStore/${dataStoreNamespace}/${key}/metaData`)
            .getData();

        return data;
```

SferaDev:
@xurxodev I've not reviewed the PR itself, but added some comments to unify style between the code on ``ConstantStorage`` and ``DataStoreStorage``

src/domain/storage/repositories/StorageClient.ts
```diff
@@ -8,6 +8,25 @@ export interface StorageClientConstructor {
     new (instance: Instance): StorageClient;
 }
 
+export type AccessRule = string;
+
+export interface SharingRule {
+    id: string;
+    access: AccessRule;
+    displayName: string;
+}
```
SferaDev, line 17: We already have an entity called ``SharingSetting`` used in the app with essentially the same behavior, I would personally re-use it (src/domain/common/entities/SharingSetting).

```ts
export interface SharingSetting {
    access: string;
    displayName: string;
    id: string;
    name?: string;
}
```

In other apps where we handle sharing settings, we even have schema validation already created for it and some helper methods to check permissions abstracted into helper files.

src/domain/common/entities/ShareableEntity.ts
```diff
@@ -0,0 +1,37 @@
...
+
+        const {
+            publicAccess = "--------",
+            userAccesses = [],
+            userGroupAccesses = [],
+            user,
+        } = this.data;
+        const token = permission === "read" ? "r" : "w";
+
+        const isUserOwner = user.id === currentUser.id;
```
SferaDev, line 19: I get a bug on some instances here, seems user can be undefined but TS does not complain. We might have some types wrong:

```ts
const isUserOwner = user?.id === currentUser.id;
```

SferaDev:
@xurxodev @adrianq I'm testing the sharing settings on constant storage and I constantly get the ``Your current dhis2 version is 2.30 and does not exist share settings for instances. This is a potencial risk!``. However I'm on 2.33 and there shouldn't be security concerns here.

I checked the code and we are checking the destination instance version to review if it's secure, but I believe we should check origin instance always and only complain if the local dataStore does not allow sharing settings.

Also I believe on constants we can disable that warning since it doesn't matter the version, sharing settings always exist.

src/data/storage/StorageDataStoreClient.ts
xurxodev: Ok, fixed!

src/data/storage/StorageDataStoreClient.ts
```diff
@@ -72,4 +72,49 @@ export class StorageDataStoreClient extends StorageClient {
...
+
+        return { ...sharing, user: { ...sharing.user, name: "" } };
+    }
+
+    public async saveObjectSharing(key: string, object: ObjectSharing): Promise<void> {
+        const metadata = await this.getMetadataByKey(key);
+
+        await this.api.post(`/sharing?type=dataStore&id=${metadata.id}`, undefined, {
+            object,
+        }).response;
```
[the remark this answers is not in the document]
xurxodev: Ok, fixed!

src/data/storage/StorageDataStoreClient.ts
xurxodev: Ok, fixed!

src/domain/common/entities/ShareableEntity.ts
xurxodev: Ok, fixed!

src/domain/storage/repositories/StorageClient.ts
xurxodev: ShareSettings is only a part of the /api/sharing network response for sharing and shareRef domain entity is not equal to network response in this case. In clean architecture is normal these cases.  I personally would not re-use it, so SharingObject and SharingRule can be moved to d2api in the future when d2api implements this feature.

xurxodev:
@adrianq fixed!

adrianq:
@xurxodev The new feature seems to be working fine. Two things I would like to check tomorrow with you:
- Existing instances should be set as private as part of the migration. If I am not wrong, new instances are set as private which is correct. However, existing instances are public.
- A bit unrelated but when editing a rule (editing not creating) the 'Download JSON' button on the last step of the sync rule generates a JSON that does not contain the latest changes.
Anyway, I am merging this and we can look into that as part of another PR

PR #850, opened 2021-03-16
[Fix] Constant storage intitialization

SferaDev:
### :pushpin: References

- https://app.clickup.com/t/fq76zq

### :memo: Implementation

- Default to ``dataStore``
- Disable cypress tests on travis

### :bookmark_tabs: Others

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,51 +6,16 @@ cache:
     directories:
         - "$HOME/.cache"
 env:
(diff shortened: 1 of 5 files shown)


adrianq (changes requested):
@SferaDev Now this works for datastore but it does not work for constant. If I have the constants set up in my DHIS2 instance and I open the app it creates a new entry in datastore and start using it as the default storage.

PR #868, opened 2021-04-07
[Feature] Include DE from program stage sections (3)

SferaDev:
### :pushpin: References

- https://app.clickup.com/t/gq3f07

### :memo: Implementation

- Include DE from program stage sections

- Tested with Malaria Adult Surveillance program (DE: tlyhOJkxzHQ)

### :bookmark_tabs: Others


adrianq (changes requested):
@SferaDev We are missing the friendly names. If you can also add the friendly name for programStages.dataEntryForms...

SferaDev:
@adrianq Updated and I just noticed that we weren't adding the dependencies to program stages and program stage sections while they're pieces of metadata that can be selected manually from the UI

PR #869, opened 2021-04-08
[Feature] Show summary during import (4)

SferaDev:
### :pushpin: References

- Related to https://app.clickup.com/t/crz38c

### :memo: Implementation

- Export ``SummaryStepContent`` and re-use it in import summary dialog

### :art: Screenshots

image
image


adrianq:
@SferaDev I will wait for @xurxodev  to review this before testing it in detail. I have just tested very quickly:
- I have exported the metadata sync rule "Admin dashboard event dataelements" from the WHO docker
- I have tried to import again and I am getting the following error:
Problem with the value of property "builder": Problem with the value of property "syncParams": One of the following problems occured: (0) Problem with property "enableMapping": it does not exist in received object {"mergeMode":"MERGE","atomicMode":"ALL","importMode":"COMMIT","importStrategy":"CREATE_AND_UPDATE","includeSharingSettings":true,"useDefaultIncludeExclude":true,"metadataIncludeExcludeRules":{}}, (1) Expected an undefined, but received an object with value {"mergeMode":"MERGE","atomicMode":"ALL","importMode":"COMMIT","importStrategy":"CREATE_AND_UPDATE","includeSharingSettings":true,"useDefaultIncludeExclude":true,"metadataIncludeExcludeRules":{}}

SferaDev:
@adrianq Makes sense... Since we never enforced schemas on the data store sync rules fail to import if they do not include [mandatory] properties that were defined AFTER the sync rules were created (in future versions).

We can discuss several options to fix this problem in a quick meeting if you want.

xurxodev (changes requested):
Thanks @SferaDev, now we have avoided the domain is coupled to the purify library and this is great.

* I have reviewed to architectural level and I have found a problem:

ReadFiles method in the RulesRepository:
I think the read files method should be in another place, maybe another concept like RulesGateway that normally Gateway is used to communications with other systems (like a disk) or another repository with another name for example FileRulesRepository where we only have one method getAll and in the future, we can have several implementations like FileJsonRulesRepository or FileXmlRulesRepository. The problem with the current implementation in the RulesRepository is that a repository should be only for a data source type and here we are mix two data sources types disk and backend. I think has more sense to separate it because a future new implementation of RulesRepository will generate a duplicate code to read files because the logic to read files is always the same.

I think move read files to another place is a simple change and we avoid problems in the future.

* I have tested the feature and simple process create sync rule, export and import seems work but we have not validated that metadataI ids, event ids, TEI ids in the sync rule exist in the instance and we can have future bugs reported by the client saying that executed imported sync rules not working. 
I think if to import a file some metadata ids, event ids, Tei ids in the sync rule does not exist in the instance we should not import the file.

To replicate export a existed sync rule and edit the file modifying any metadata ids, event ids, or Tei ids, import it and execute the sync rule.

SferaDev:
- ReadFiles method in the RulesRepository

Yeah. However since the read zip file and get json files could be potentially used on packages/modules (we already have import there and similar code already exists) wouldn't it make sense to completely decouple it from SyncRule entity?

- Validation of UIDs

We kind of discussed this with Adrian, and we removed that kind of validations from the current scope of this feature. We definetely wanted to have that kind of additional checks but it was impossible in the available time.

xurxodev:
Yes! if using generics parameters we can create an agnostic FileDataRepository using purify to decode any type I think is good.

SferaDev:
> Yes! if using generics parameters we can create an agnostic FileDataRepository using purify to decode any type I think is good.

Created a new ``FileRepository`` -> ``FileDataRepository`` (and renamed the existing one to ``InstanceFileRepository`` as the original naming was too broad).

@xurxodev Please check it out and verify it is more reusable now.

I preferred to keep the decoding with purify to the ``RulesRepository`` though to split responsibilities.

xurxodev:
@SferaDev With the current solution in the FileRepository, we now avoid duplicate code to load files from different entities as Rules, Modules etc.. but if we create a new implementation of RulesRepository, we still will generate a duplicate code for readFiles method. I think if we create RulesFileRepository or a similar name with the current readFiles method avoid all duplicate code problems

SferaDev:
> @SferaDev With the current solution in the FileRepository, we now avoid duplicate code to load files from different entities as Rules, Modules etc.. but if we create a new implementation of RulesRepository, we still will generate a duplicate code for readFiles method. I think if we create RulesFileRepository or a similar name with the current readFiles method avoid all duplicate code problems

@xurxodev Not sure when we will create a new implementation of the RulesRepository (if we ditch d2-api and use a new API connection library?). To me it's a very unlikely case and doesn't weigh enough to split the RulesRepository into multiple smaller repositories (consisting of one single method).

xurxodev:
> > @SferaDev With the current solution in the FileRepository, we now avoid duplicate code to load files from different entities as Rules, Modules etc.. but if we create a new implementation of RulesRepository, we still will generate a duplicate code for readFiles method. I think if we create RulesFileRepository or a similar name with the current readFiles method avoid all duplicate code problems
> 
> @xurxodev Not sure when we will create a new implementation of the RulesRepository (if we ditch d2-api and use a new API connection library? And not only ditch but assume that we have multiple coexisting API connection libraries, note that this does not refer to different data source backends but to where the app runs which is against a DHIS2 always per definition both in browser and cli). To me it's a very unlikely case and doesn't weigh enough to split the RulesRepository into multiple smaller repositories (consisting of one single method).

I do not know if we will generate a new implementation of RulesRepository but if something can generate duplicate code it is a clue that it should not be there

xurxodev (approved):

PR #902, opened 2021-10-18
[Feature] Update dependencies and add proxy

SferaDev:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/1h8226v

### :memo: Implementation

- Update dependencies (all except for cryptr)
- Fix lint issues


xurxodev (changes requested):
I have found an error to click on the top left logo in the toolbar, try to navigate to the http://localhost:8081/dhis2 URL and fail

SferaDev:
> I have found an error to click on the top left logo in the toolbar, try to navigate to the http://localhost:8081/dhis2 URL and fail

Since this only happens in development, I would consider it a minor annuance. If you want to take a look and fix it go ahead, if not we can take a look in the future.

adrianq:
> > I have found an error to click on the top left logo in the toolbar, try to navigate to the http://localhost:8081/dhis2 URL and fail
> 
> Since this only happens in development, I would consider it a minor annuance. If you want to take a look and fix it go ahead, if not we can take a look in the future.

Yes, I think it is a very minor detail. I wouldn't spend any time on fixing this issue

PR #916, opened 2021-11-17
[Feature] DE analytics mapping

SferaDev:
### :memo: Implementation

- [x] Create new mapping model -  program with data elements to aggregated
- [x] Modify sync rule validation to TEI, event or aggregation
- [x] Update aggregated mapper to support mapping from program De to Aggregated DE during the sync
- [x] Add new aggregated mapper tests
- [x] Avoid duplicate dimension error if exists duplicate data elements by program
- [x] Rename summary type for aggregations in events	

Due to a limitation in the analytics endpoint, it's not possible to request with dx dimension by program stage and data element. Only program and data element is allowed.

For this reason, in the mapping configuration, the new model uses the complex id program-dataElement.

Using this complex Id is not possible to map the same data element in different program stages of the same program to different aggregated data elements. 

Jira-issue: https://jira.dhis2.org/browse/DHIS2-12382


SferaDev (commented):
Tested and works fine!

Thanks @xurxodev

PR #921, opened 2022-01-28
Feature/extend dhis2 version compatibility

xurxodev:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/1uq6fme

### :memo: Implementation

- [x] Create new action for parents in package list page
- [x] Create dialog to extends dhis2 compatibility
- [x] Create a new use case to create n packages from existing packages with the same module and version 
- [x] Cherry-pick to bring units tests and CI fixed

### :art: Screenshots

Captura de pantalla 2022-01-28 a las 12 43 19
Captura de pantalla 2022-01-28 a las 12 44 03

### :bookmark_tabs: Others


ifoche (commented):
Sorry @xurxodev but there is a last-minute requirement from the client and I would like to check with you. The client is requesting now, that even if we don't touch the rest of the package that is inside the json file, we still apply the transformations that we do, when we are in one version and we generate a package from another version. Like if you are doing the package extension from 2.34 to 2.37, the resulting package for 2.37 should have the transformations applied that turn that 2.34 metadata suitable for 2.37. 

Would that be possible from the code point of view?

PR #935, opened 2022-11-07
Convert EFH-Sync from widget to app

tokland:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/3rchac7, https://app.clickup.com/t/3megcwe

### :memo: Implementation

- Use type "app" instead of "widget"
- Add route / for the EFH sync buttons page
- Add route /dashboard to access mdsync normal landing page.
- IMPORTANT, to validate with @xurxodev: So the user can download the JSON response from the UI, I've added `response?: object` to `SynchronizationResult`. It's typed as `object` because the entity does not need to know it's structure (that's data layer implementation)

### Screenshots

image

image


tokland:
- [ ] [tracker capture iframe] Hide event notes
- [ ] Add JSON response link
- [ ] Beautify stats (use table)

adrianq:
@tokland I am merging this as we are already using it but feel free to touch base with @xurxodev on the clean arch question

PR #957, opened 2024-06-06
remove sharing objects

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8694rk79p

### :memo: Implementation

- [x] exclude sharing object from metadata if switch is selected

### :bookmark_tabs: Others

#8694rk79p


ifoche:
Task linked: [CU-8694rk79p MDsync exports sharing information even when it shouldn't](https://app.clickup.com/t/8694rk79p)

MiquelAdell (approved):

tokland (approved):
Looks good to me, `sharing` unifies the from user+userAccesses+userGroupAccesses, so conceptually it makes sense to consider it a userProperty (that can be later removed if includeSharingSettings = false, for example).

PR #959, opened 2024-06-08
exclude default categories from dataElements

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8694rjjmk

### :memo: Implementation

In DHIS 2.38.6 the defaults parameter is not working in the `metadata` endpoint
`/api/metadata.json?fields=id&filter=id:eq:data_element_id&categoryCombo&defaults=INCLUDE/EXCLUDE;`
so I'm manually removing the default `categoryCombo` from `dataElements/dataSets`

image

### :bookmark_tabs: Others

#8694rjjmk


ifoche:
Task linked: [CU-8694rjjmk MDSync is generating the default category](https://app.clickup.com/t/8694rjjmk)

MiquelAdell (approved):

src/domain/metadata/usecases/MetadataSyncUseCase.ts
```diff
@@ -41,18 +49,27 @@ export class MetadataSyncUseCase extends GenericSyncUseCase {
...
             for (const element of elements) {
                 //ProgramRules is not included in programs items in the response by the dhis2 API
                 //we request it manually and insert it in the element
                 const fixedElement =
                     type === "programs" ? await this.requestAndIncludeProgramRules(element as Program) : element;
 
+                // In DHIS 2.38.5 and above the defaults parameter is not working
+                // /api/metadata.json?fields=id&filter=id:eq:data_element_id&categoryCombo&defaults=INCLUDE/EXCLUDE; this is not working
+                // so manually removing the default categoryCombo
+                const elementWithoutDefaults =
```
tokland, line 63: A use case should not make any reference to DHIS2 (that's an implementation detail). If DHIS2 has a bug (Indeed, I've checked it: 2.36 OK, 2.38 FAIL), any extra processing should be done in the data layer, probably the metadata DHIS2 repositoriy (or any other file which calls the API with that option `defaults: boolean`  -> I've found also utils/synchronization.ts)

Note: checking dhis2-core, DefaultCategoryService.java, the models that have a default object are the 4 expected: category, categoryOption, categoryCombo, categoryOptionCombo.

tokland (changes requested):

src/utils/d2-utils.ts
```diff
@@ -22,3 +25,47 @@ export function getD2APiFromInstance(instance: Instance) {
     */
     return new D2Api({ baseUrl: instance.url, auth: instance.auth, backend: "fetch" });
 }
+
```
tokland, line 28: Let's move the file to data/ so it's more CleanArch compliant.

tokland (approved):
[code-only review] All good. I'd just move the file (not a major issue)

MiquelAdell:
I guess @eperedo commit fixes @tokland request, does it? I've re-requested the review from Arnau

tokland (approved):
Nothing more to add, it's ok by me now.

PR #960, opened 2024-06-10
[fix] Security issues

tokland:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8694rzyqd

### :memo: Implementation

- [x] Fix reports from `snyk code test` (src/)
- [x] Fix reports from `snyk test` (d2-api dependencies)
- [x] Fix reports from `snyk test` (d2-ui-components dependencies)
- [x] Fix reports from `snyk test` (feedback-component dependencies)
- [ ] Fix reports from `snyk test` (non-EST dependencies) -> if possible
- [x] Replace react-json-view (with security issues, unmantained) to react-json-view-lite
- [ ] Setup Dependency-Track


tokland:
I am moving the Slack conversation here. @MiquelAdell reported: 

```
←→1 of 4 errors on the page
Unhandled Rejection (Error): Unsupported state or unable to authenticate data
▶ 3 stack frames were collapsed.
InstanceD2ApiRepository.decryptPassword
src/data/instance/InstanceD2ApiRepository.ts:123
  120 | }
  121 | 
  122 | private decryptPassword(password?: string): string {
> 123 |     return password ? new Cryptr(this.encryptionKey).decrypt(password) : "";
      | ^  124 | }
```

tokland:
I could reproduce the problem (https://dev.eyeseetea.com/who-dev-238/)

I tried to upgrade to latest cryptr version (6.3.0), and now I get another error ("list" argument must be an Array of Buffers), which comes from an extremely old dependency of cryptr: browserify-aes. 

The bug is not simple to debug, and I am not sure we can solve it as it's down the dependency chain. I'll need more time for that.

If we need to merge this now, I'd revert to the old known version working (4.0.2), the snyk warning was: "Insufficient Entropy [Medium Severity]"

MiquelAdell:
@tokland reverting cryptr to 4.0.2 would leave this part of the security fix "as if it was not worked on" wouldn't it? just to make a better informed decision

tokland:
> @tokland reverting cryptr to 4.0.2 would leave this part of the security fix "as if it was not worked on" wouldn't it? just to make a better informed decision

Yes, we'll have again that "medium severity" warning by snyk.

tokland:
- [ ] Revert cryptr@4.0.2 to avoid exception on decryp

MiquelAdell (approved):
I'm approving the PR even if there is some additional work that we'll need to perform

tokland:
@MiquelAdell About cryptr, I think the best solution would be to replace it altogether and use a more modern package in npm. Of course, your dev should check that the encode/decode still give the same results so existing values can be decoded.

PR #961, opened 2024-06-25
use default app namespace

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8694ubx3x

### :memo: Implementation
- [x] Update i18n types
- [x] Set default namespace in `app-context.ts`

### :bookmark_tabs: Others

#8694ubx3x


tokland (approved):
Some context: This small hack (`setDefaultNamespace`) is necessary because other packages may change it (d2-ui-components, feedback-component, some d2 package). If it works, let's go ahead it. Let's keep the idea for future projects: we should use a single, wrapper `t`, where we put the explicit namespace.

MiquelAdell:
Just context for the reason of the delay. This failed for some reason. We were waiting for some help in fixing the issue from @eperedo but did not got to it. It's in halt wating for some senior help

eperedo:
The problem is that sometimes the i18n namespace from other libraries (d2-ui-components, feedback-component) is replacing the app namespace. After that i18n is looking for translations in the libraries instead the app. My guess is that the problem could be related to the feedback-component repo. We should explicit set the namespace similar what we have in the [skeleton app](https://github.com/EyeSeeTea/dhis2-app-skeleton/commit/54fcc221503d23f3bf4490d4315679b4cf60034b) or [d2-ui-components](https://github.com/EyeSeeTea/d2-ui-components/blob/development/src/utils/i18n.ts) publish a beta version and test here.

ifoche:
Task linked: [CU-8694ubx3x check all apps for conflicts between translations and feedback component](https://app.clickup.com/t/8694ubx3x)

PR #963, opened 2024-07-01
Sync all orgunits toggle for metadata rules

p3rcypj:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8694rb1fr

### :memo: Implementation

- Added button to sync all OUs dynamically, regardless of id/whether they are created/or removed.
- EFH buttons can now execute multiple rules.
- Almost 90% code done to allow "Sync all" metadata types feature. Not done in this PR due to a lack of checking that it really works for all the metadata types. Doing this feature only for "organisationUnits" would cost much more time because would have to constantly check for "organisationUnits" collection name rather than just using an array for the types we want to "sync all" and add them on the requests made.

On both DHIS2 images should work:

```
docker.eyeseetea.com/samaritans_efh/dhis2-data:2.36.11-pablo-test-2
docker.eyeseetea.com/samaritans_efh/dhis2-data:2.36.11-sp-efh-skeleton
```

Should create a `Get organisation units` rule first (rules are executed in order by name) and code `EFH_METADATA_ORGUNITS`

image

In order to continue to step 2 (Select metadata) check you already tested connection with "platform" source instance in "Instance settings". On step 2, select Organisation Unit metadata type from the dropdown and check the "Sync all organisation units" toggle.

image

On step 4: remove all the dependencies for Organisation Units

image

On step 5: deselect "Include owner and sharing settings" and select "Remove users and references (UID)"

image

Here you have the summary for all steps:

image

After that, go to `/efh` and execute "Get configuration from HQ server". It should execute first the org units rule and later the previous one we had already. It will show two reports.

**DISCLAIMER: Don't execute "PUSH DATA TO HQ SERVER" without changing the target destination instance to another one of your own.**

### :bookmark_tabs: Others

-   Any change in the [GUI library](https://github.com/EyeSeeTea/d2-ui-components)? If so, what branch/PR? **No**
-   Any change in the [D2 Api](https://github.com/EyeSeeTea/d2-api)? If so, what branch/PR? **No**


src/presentation/react/core/components/metadata-table/MetadataTable.tsx
```diff
@@ -251,6 +264,10 @@ const MetadataTable: React.FC<MetadataTableProps> = ({
         updateFilters({ disableFilterRows });
     };
 
+    const changeOrgUnitsSyncAll = (value: boolean) => {
+        notifyModelSyncAllChange(value);
+    };
```
tokland, line 269: Unless you have some specific idea in mind, you can just use notifyModelSyncAllChange, as there is no transformation here.

src/presentation/react/core/components/sync-wizard/common/MetadataSelectionStep.tsx
```diff
@@ -122,10 +124,46 @@ export default function MetadataSelectionStep({ syncRule, onChange }: SyncWizard
...
+        if (!_.isEmpty(metadataSyncAll)) {
+            compositionRoot.metadata.getByIds(syncRule.metadataIds, remoteInstance, "id").then(metadata => {
+                const idsFromSyncAllMetadataTypes = _.compact(
+                    _(metadata)
+                        .pick(metadataSyncAll)
+                        .values()
+                        .value()
+                        .flat()
+                        .map(entity => entity?.id)
+                );
```
tokland, line 137: same

src/presentation/react/core/components/sync-wizard/common/MetadataSelectionStep.tsx
```diff
@@ -122,10 +124,46 @@ export default function MetadataSelectionStep({ syncRule, onChange }: SyncWizard
...
+                onChange(syncRule.updateMetadataTypes(ruleTypes).updateMetadataSyncAll(syncAllTypes));
+
+                return syncAllTypes;
+            });
+        },
+        [model, syncRule, onChange]
+    );
+
+    const modelIsSyncAll = useMemo(() => metadataSyncAll.includes(model.getCollectionName()), [metadataSyncAll, model]);
+
```
tokland, line 166: [no need to change] Even when arriving at components with all useEffect/useCallback in-line, it's a good practice to create the new code as hooks, so we go in the right direction. Specially important for inline useEffect, which are nameless blocks of actions.

src/presentation/react/core/components/sync-wizard/common/SummaryStep.tsx
```diff
@@ -183,8 +184,19 @@ export const SummaryStepContent = (props: SummaryStepContentProps) => {
...
+                        syncRule.metadataSyncAll.map(
+                            async type =>
+                                await compositionRoot.metadata
+                                    .listAll({ type: type as keyof MetadataEntities, fields: { id: true } }, instance)
+                                    .then(metadata => ({
+                                        [type]: metadata.filter(({ id }) => !syncRule.excludedIds.includes(id)).length,
+                                    }))
+                        )
+                    ).then(typeLength => setSyncAllTypesLength(Object.assign({}, ...typeLength)));
+
```
tokland, line 197: As a general rule, we should have a useCase which does the task so you don't do this Promise.all+map+... in the view.

src/presentation/react/core/components/sync-wizard/common/SummaryStep.tsx
```diff
@@ -183,8 +184,19 @@ export const SummaryStepContent = (props: SummaryStepContentProps) => {
...
                     //type is required to transform visualizations to charts and report tables
-                    compositionRoot.metadata.getByIds(ids, instance, "id,name,type").then(updateMetadata);
+                    compositionRoot.metadata.getByIds(ids, instance, "id,name,type").then(updateMetadata).then();
```
tokland, line 199: empty `then()`?

src/presentation/react/core/components/sync-wizard/common/SummaryStep.tsx
```diff
@@ -216,32 +228,49 @@ export const SummaryStepContent = (props: SummaryStepContentProps) => {
...
+                .keys()
+                .concat(syncRule.metadataSyncAll)
+                .uniq()
+                .sort()
+                .value()
+                .map(metadataType => {
+                    //@ts-ignore
+                    const modelByMetadataType = api.models[metadataType];
+                    if (!modelByMetadataType) {
+                        console.warn(`Metadata type "${metadataType}" not supported in d2-api`);
```
tokland, line 241: Not important here, it's just a message, but we won't like to find references to d2-api (data layer) in the presentation layer.

src/presentation/react/core/components/sync-wizard/common/SummaryStep.tsx
```diff
@@ -216,32 +228,49 @@ export const SummaryStepContent = (props: SummaryStepContentProps) => {
...
-                }
-                const itemsByType = metadata[metadataType as keyof MetadataEntities] || [];
+            {_(metadata)
+                .keys()
+                .concat(syncRule.metadataSyncAll)
+                .uniq()
+                .sort()
+                .value()
+                .map(metadataType => {
+                    //@ts-ignore
```
tokland, line 238: It's better to a casting `as keyof MetadataEntities` than ts-ignore. Of course, the better would be not to have neither :)

src/presentation/react/core/components/sync-wizard/metadata/MetadataIncludeExcludeStep.tsx
```diff
@@ -44,7 +44,13 @@ const MetadataIncludeExcludeStep: React.FC<SyncWizardStepProps> = ({ syncRule, o
...
                         .getByIds(syncRule.metadataIds, instance, "id,name,type") //type is required to transform visualizations to charts and report tables
                         .then((metadata: MetadataPackage<MetadataEntity>) => {
-                            const models = _.keys(metadata).map((type: string) => modelFactory(type));
+                            const models = _(metadata)
+                                .keys()
+                                .concat(syncRule.metadataSyncAll)
+                                .sort()
+                                .uniq()
+                                .value()
+                                .map((type: string) => modelFactory(type));
```
tokland, line 53: Why `: string`?

src/presentation/react/core/components/sync-wizard/metadata/MetadataIncludeExcludeStep.tsx
```diff
@@ -44,7 +44,13 @@ const MetadataIncludeExcludeStep: React.FC<SyncWizardStepProps> = ({ syncRule, o
...
 
                             const options = models
                                 .filter((model: typeof D2Model) => model.getMetadataType() !== defaultName)
```
tokland, line 56: same. TS should know the type.

src/presentation/react/core/components/toggle/Toggle.tsx
```diff
@@ -3,7 +3,7 @@ import _ from "lodash";
 
 interface InputParameters {
     disabled?: boolean;
-    label: string;
+    label: string | React.ReactNode;
```
tokland, line 6: A string is also a React.ReactNode (AFIK)

src/presentation/webapp/sp-emergency-responses/EmergencyResponsesSyncHomePage.tsx
```diff
@@ -27,12 +26,46 @@ export const EmergencyResponsesSyncHomePage: React.FC<EmergencyResponsesSyncHome
     const rules = useSyncRulesList(emergencyType);
     const [isRunning, runSyncRule] = useSyncRulesExecuter({ emergencyType, logs });
 
+    const executeRules = React.useCallback(
+        (rules: SynchronizationRule[]) =>
+            _.orderBy(rules, rule => rule.code === "EFH_METADATA_ORGUNITS", "desc").forEach(rule => {
+                console.debug("Running: " + rule.name);
+                runSyncRule(rule);
```
tokland, line 33: runSyncRule is an async function. Being called in a forEach, that will call them in parallel. Probably not what we want. Adding await it's not sufficient, we'll need to convert the forEach to a for-loop, or probably better, use `promiseMap`.

src/presentation/webapp/sp-emergency-responses/EmergencyResponsesSyncHomePage.tsx
```diff
@@ -27,12 +26,46 @@ export const EmergencyResponsesSyncHomePage: React.FC<EmergencyResponsesSyncHome
...
+
+    const metadataRules = React.useMemo(() => {
+        const metadataRules = rules.filter(rule => rule.type === "metadata");
+        return { rules: rules, execute: () => executeRules(metadataRules) };
+    }, [rules, executeRules]);
+
+    const eventsRules = React.useMemo(() => {
+        const eventsRules = rules.filter(rule => rule.type === "events");
+        return { rules: rules, execute: () => executeRules(eventsRules) };
+    }, [rules, executeRules]);
```
tokland, line 46: Not sure about these 2 structures. The "rules" property seems unused. And the execute would seem to be in fact a useCallback.

```typescript
const executeEventsRules = React.useCallback(() => {
    const eventsRules = rules.filter(rule => rule.type === "events");
    executeRules(eventsRules);
}, [rules, executeRules]);
```

src/presentation/react/core/components/metadata-table/MetadataTable.tsx
p3rcypj: I will leave a comment but it would be for a future feature where every metadataType can be "sync all" activated

src/presentation/react/core/components/sync-wizard/metadata/MetadataIncludeExcludeStep.tsx
p3rcypj: not sure why it was there...

src/presentation/react/core/components/toggle/Toggle.tsx
p3rcypj: Didn't know about it :o well, it make sense at the end of the day

tokland (approved):
[code-only review] all good by me!

MiquelAdell (approved):
It works from the front side. @tokland do you want to review the changes?

tokland (approved):
[code-only review] all good

PR #977, opened 2024-12-03
fix: DHIS2 versions drop downs in package generation/extension

MatiasArriola:
### :pushpin: References

-   **Issue:** Closes #? https://app.clickup.com/t/8696q81xa #8696q81xa

### :memo: Implementation

- Moved the versions arrays from `PackagesExtendCompatibilityDialog` and `CreatePackageFromFileDialog` to `DhisReleasesLocalRepository`. The list is still hardcoded but newer versions were added, and are centralized in one place.
- I was not sure about adding `DhisRelease` as an entity because it is just a string, but we could change that type in the future for something more robust.
- An alternative (to avoid changing this array on every new dhis release) could be to prefetch the list of releases at build time from https://releases.dhis2.org/v1/versions/stable.json and use another Repository implementation to return them. But maybe the manual update is still preferred for more control on what versions to show.

https://github.com/user-attachments/assets/06e2aa54-d9b7-481f-b351-8f5058c43b9c

(mouse pointer not showing for some reason. I will fix it for next PRs)

### :bookmark_tabs: Others


ifoche:
Task linked: [CU-8696q81xa Check DHIS2 versions drop downs in package generation](https://app.clickup.com/t/8696q81xa)

src/presentation/react/core/components/create-package-from-file-dialog/CreatePackageFromFileDialog.tsx
```diff
@@ -57,6 +58,7 @@ export const CreatePackageFromFileDialog: React.FC<CreatePackageFromFileDialogPr
 
     useEffect(() => {
         compositionRoot.user.current().then(({ userGroups }) => setUserGroups(userGroups));
+        compositionRoot.dhisReleases.getSupportedDhisVersions().then(versions => setDhisSupportedVersions(versions));
```
eperedo, line 61: Not a big deal, but since we're using this method here and in `PackagesExtendCompatibilityDialog.tsx` maybe we can extract it into a custom hook:

```ts

export function useGetSupportedVersion() {
    React.useEffect(() => {
        compositionRoot.dhisReleases.getSupportedDhisVersions().then(setDhisSupportedVersions);
    }, [compositionRoot]);
}

```

this way if we need to fetch the versions in other components we can reuse the same hook. What do you think?

eperedo (changes requested):
[code only review]
Looks good! Just a minor suggestion about reusing a method.

MatiasArriola:
Thanks for your review @eperedo . Good catch on the hook! I've pushed the suggested changes, let me know what you think

eperedo (approved):
All good, thanks @MatiasArriola!

MiquelAdell (approved):

PR #986, opened 2025-02-13
Remove slash from call to API

adrianq:
### :pushpin: References

-   **Issue:** Closes #?

### :memo: Implementation

-

### :bookmark_tabs: Others


xurxodev (changes requested):
thanks @adrianq 

This change introduce an error in development
Screenshot 2025-02-18 at 10 53 45

i18n/en.pot
```diff
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2025-02-04T09:15:52.687Z\n"
-"PO-Revision-Date: 2025-02-04T09:15:52.687Z\n"
+"POT-Creation-Date: 2025-02-19T09:55:28.790Z\n"
+"PO-Revision-Date: 2025-02-19T09:55:28.790Z\n"
```
tokland, line 9: No need to commit. Not sure why, mdsync seems to be the only project that updates the timestamp when no changes actually exist.

src/data/role/RoleD2ApiRepository.ts
```diff
@@ -0,0 +1,39 @@
+import { Role } from "../../domain/role/Role";
+import { RoleRepository } from "../../domain/role/RoleRepository";
+import { D2Api } from "../../types/d2-api";
+
+export class RoleD2ApiRepository implements RoleRepository {
+    constructor(private api: D2Api) {}
+    async getByName(name: string): Promise<Role | undefined> {
```
tokland, line 7: blank line

src/data/role/RoleD2ApiRepository.ts
```diff
@@ -0,0 +1,39 @@
...
+        const { userRoles } = await this.api.metadata
+            .get({
+                userRoles: {
+                    fields: { id: true, name: true, publicAccess: true, description: true },
+                    filter: { displayName: { eq: name } },
```
tokland, line 12: Not sure, just to confirm, the filter we need is over the displayName (translated) or name?

src/data/role/RoleD2ApiRepository.ts
```diff
@@ -0,0 +1,39 @@
...
+                    userRoles: true,
+                },
+            })
+            .getData();
+
+        if (userRoles && userRoles.length > 0) {
```
tokland, line 18: userRoles should be an array, no need to for the first check

src/data/role/RoleD2ApiRepository.ts
```diff
@@ -0,0 +1,39 @@
...
+            const role = userRoles[0];
```
tokland, line 19: The most idiomatic and type-safe approach is to first declare `const role = userRoles[0];` and then check `if (role) { ... }`.  

In fact, TS isn't reporting an error because `"noUncheckedIndexedAccess": false` is set in `tsconfig.json`, which we shouldn't have! However, changing it now is out of scope, as it would require adjusting **226 occurrences**, according to `npx tsc`.  

@9sneha-n, could you add this to the refactoring notes if it's not already there? It's definitely something to tackle in the future.

src/data/role/RoleD2ApiRepository.ts
```diff
@@ -0,0 +1,39 @@
...
+
+            return Role.create({
+                id: role.id,
+                name: role.name,
+                description: role.description,
+                publicAccess: role.publicAccess,
+            });
```
tokland, line 26: When no transformation is needed we may simply pass `role`, but it's also ok be explicit.

src/domain/role/AppRoles.ts
```diff
@@ -0,0 +1,24 @@
+export const AppRoles: {
+    [key: string]: {
+        name: string;
+        description: string;
+        initialize: boolean;
+    };
+} = {
```
tokland, line 7: This is moved code, but as before, having `"noUncheckedIndexedAccess": true` is masking a subpar approach. The way we are defining the type makes it type-unsafe and hinders code navigation. There are two approaches:  

- Using `satisfies Record<string, { name: string; description: string; initialize: boolean; }>` but that requires TS 4.9 (we have 4.4).  
- Defining a type alias + builder and letting TS infer the final type:  

  ```typescript
  type AppRole = { name: string; description: string; initialize: boolean };

  function role(role: AppRole): AppRole {
      return role;
  }
  ```

  And then use it:  
  ```typescript
  CONFIGURATION_ACCESS: role({ ... })
  ```

src/domain/role/ValidateRolesUseCase.ts
```diff
@@ -0,0 +1,24 @@
...
+    constructor(private roleRepository: RoleRepository) {}
+
+    public async execute(): Promise<void> {
+        for (const role in AppRoles) {
+            const { name, description, initialize } = AppRoles[role];
+
+            if (initialize) {
+                const role = await this.roleRepository.getByName(name);
+
+                if (role === undefined) {
```
tokland, line 16: more idiomatic: `if (!role)`

tokland (commented):
In-line comments:

src/data/role/RoleD2ApiRepository.ts
xurxodev: :) No worries, fixed

xurxodev:
@tokland All changes is done

tokland (approved):
Just a minor thing, the explicit type can now be removed, approving.

xurxodev (approved):

PR #993, opened 2025-03-04
refactor settings parse 1 and some common code refactor

9sneha-n:
### :pushpin: References

-   **Issue:** Closes #86980yb2n? https://app.clickup.com/t/86980yb2n, 

### :memo: Implementation
1. Use clean architecture - refactor entities, usecases, repos
2. Create a new composition root using latest coding standards
3. Remove implements UseCase in UseCases, this creates pain point in navigation

-

### :bookmark_tabs: Others


ifoche:
Task linked: [CU-86980yb2n Refactor settings](https://app.clickup.com/t/86980yb2n)

xurxodev (changes requested):
thanks @9sneha-n 

In this PR there are two principal changes:
## Rename oldConfigAppRepository to StorageClientD2Repository

I think the name is better now. However we can to use this PR to improve the architecture.

In clean architecture, the owner of a repository is the use case. The use case decide the api of the interface according its necessity.

For this reason GetStorageClientUseCase and SetStorageConfigUseCase has access to getStorageClient, getUserStorageClient, changeStorageClient methods of the repository.

However there are multiple repositories where StorageClientRepository is injected and this is incorrect for several reasons:
* A repository invoke other repository (code smell)
* Other repositories has access to methods for they are not owners and this a risk. For example: changeStorageClient

There are several solutions, the most simple is to create a new interface named for example StorageClientFactory with the unique method used in other repositories, getStorageClient. Then StorageClientD2Repository implements also this new interface. And this new interface is injected in other repositories.

```TypeScript
// Data layer
export interface StorageClientFactory {
    getStorageClient(): Promise<StorageClient>
}

export class StorageClientD2Repository implements StorageClientRepository, StorageClientFactory {
...
}

// used in other repositories: example
export class CustomDataD2ApiRepository implements CustomDataRepository {
    constructor(private storageClientFactory: StorageClientFactory) {}

    async get<T extends CustomData>(key: string): Promise<T | undefined> {
        const storageClient = await this.getStorageClient();
        return storageClient.getObject<T>(key);
    }
    async save<T extends CustomData>(key: string, data: T): Promise<void> {
        const storageClient = await this.getStorageClient();
        await storageClient.saveObject<T>(key, data);
    }

    private getStorageClient(): Promise<StorageClient> {
        return this.storageClientFactory.getStorageClient();
    }
}
```

With this solution we solve:
* A repository invoke other repository (code smell)
* Other repositories has access to methods for they are not owners and this a risk. For example: changeStorageClient

## Create a new composition root
I think that is a good idea create a new composition root equal to other projects. And I think is good that it growth step by step staring with StorageClient use cases.

We are going to have the two composition Roots in parallel a long time.

This is a technique of refactoring called Parallel Change: https://xurxodev.com/trabajando-con-codigo-legado-parallel-change/ (sorry its in Spanish but you can translate the page)

It would be recommendable mark the old composition root as deprecated explaining that there are a new composition root. In this way it's clear for all people when need create something new to use the new compositions root.

And my last recommendation is in getTestCompositionRoot. I think would be great receive by arguments the repositories instead of hardcoded it. With this approach, every test of a page can decide pass specific repositories and using for example ts-mockito or fake manual repository.

9sneha-n:
Thanks @xurxodev. I have addressed your comments, ready for re-review .

xurxodev (approved):
thanks @9sneha-n 

There is an architecture problem now, but it's not the problem of this PR.

Now parts of domain code contains coupling to data artifacts (see the imports). Example: 
```typescript
import { ConfigRepository } from "../../config/repositories/ConfigRepository";
import { StorageClientFactory } from "../../../data/config/StorageClientFactory";
import { CustomData } from "../entities/CustomData";

export interface CustomDataRepositoryConstructor {
    new (configRepository: ConfigRepository): CustomDataRepository;
    new (storageClientFactory: StorageClientFactory): CustomDataRepository;
}

export interface CustomDataRepository {
```

But the problem is to define the constructor of an implementation of a repository in domain.  That is not correct.

Maybe interesting write down to refactor in the future.

9sneha-n:
> thanks @9sneha-n
> 
> There is an architecture problem now, but it's not the problem of this PR.
> 
> Now parts of domain code contains coupling to data artifacts (see the imports). Example:
> 
> ```ts
> import { ConfigRepository } from "../../config/repositories/ConfigRepository";
> import { StorageClientFactory } from "../../../data/config/StorageClientFactory";
> import { CustomData } from "../entities/CustomData";
> 
> export interface CustomDataRepositoryConstructor {
>     new (configRepository: ConfigRepository): CustomDataRepository;
>     new (storageClientFactory: StorageClientFactory): CustomDataRepository;
> }
> 
> export interface CustomDataRepository {
> ```
> 
> But the problem is to define the constructor of an implementation of a repository in domain. That is not correct.
> 
> Maybe interesting write down to refactor in the future.

Thanks @xurxodev . I will note this and ask @adrianq to create a task for this refactor.

adrianq:
Done https://app.clickup.com/t/8698brwj9

PR #1012, opened 2025-06-09
Fix: Show also tei sync report if rule has any TEI selected

anagperal:
### :pushpin: References

-   **Issue:** Closes #8699bn9tq [Metadata Sync - Debug & Resolve Issues for Setup](https://app.clickup.com/t/8699bn9tq)

### :memo: Implementation
- Show also tei sync report if rule has any TEI selected
- Fix error with `orgUnitPath`: Send only allowed params when POST TEIs

- Show also tei sync report if rule has any TEI selected

BEFORE:
Screenshot from 2025-06-09 15-38-36

AFTER:
Screenshot from 2025-06-09 15-38-23

- Fix error with `orgUnitPath`: Send only allowed params when POST TEIs

BEFORE:
Screenshot from 2025-06-12 11-52-22

AFTER:
Screenshot from 2025-06-12 11-50-40

### :bookmark_tabs: Others

### DIFF
--- a/src/data/events/EventsD2ApiRepository.ts
+++ b/src/data/events/EventsD2ApiRepository.ts
@@ -10,10 +10,7 @@ import { EventsPackage } from "../../domain/events/entities/EventsPackage";
 import { ProgramEvent } from "../../domain/events/entities/ProgramEvent";
 import { EventsRepository } from "../../domain/events/repositories/EventsRepository";
 import { Instance } from "../../domain/instance/entities/Instance";
-import {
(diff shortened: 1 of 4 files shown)


xurxodev (changes requested):
Thanks @anagperal,

It works fine.

I have a few minor change to request.

Currently, the validation for showing an empty summary is in a different layer for events and TEI. I think it’s better to have it in the use case layer, like it's done for TEI.

Could you move the validation to the use case for events as well?

PR #1014, opened 2025-06-11
[fix] upgrade dependencies - security issues

MatiasArriola:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8694rzyqd

### :memo: Implementation

- [ ] Fix issues reported by `snyk test`
  - [x] upgrade `@eyeseetea/d2-ui-components`, `@eyeseetea/feedback-component`
  - [x] upgrade `@material-ui`, `axios`
  - [x] upgrade `@dhis2/app-runtime`, `typescript`
  - [ ] upgrade to `@octokit/rest@21.0.0`
  - [ ] replace `cryptr` with other implementation
  - [ ] upgrade `@eyeseetea/d2-api` (breaking changes, evaluate impact)
- [ ] Fix `yarn audit issues`
- [ ] Fix Dependency-Track issues
- [ ] Upgrade Node.JS version?


tokland:
@MatiasArriola For the unticked tasks, can you add a comment like "NOT-DONE - ...reason..."? This way, you can mark all the tasks as completed; otherwise it will still appear that some work is pending.

(also, check conflicts with development)

README.md
```diff
@@ -54,9 +54,9 @@ Notes:
 
 -   Requests to DHIS2 will be transparently proxied (see `src/setupProxy.js`) from `http://localhost:8081/dhis2/path` to `http://localhost:8080/path` to avoid CORS and cross-domain problems.
 
--   The optional environment variable `REACT_APP_DHIS2_AUTH=USERNAME:PASSWORD` forces some credentials to be used by the proxy. This variable is usually not set, so the app has the same user logged in at `REACT_APP_DHIS2_BASE_URL`.
+-   The optional environment variable `VITE_DHIS2_AUTH=USERNAME:PASSWORD` forces some credentials to be used by the proxy. This variable is usually not set, so the app has the same user logged in at `VITE_DHIS2_BASE_URL`.
 
--   The optional environment variable `REACT_APP_PROXY_LOG_LEVEL` can be helpful to debug the proxyfied requests (accepts: "warn" | "debug" | "info" | "error" | "silent")
+-   The optional environment variable `VITE_PROXY_LOG_LEVEL` can be helpful to debug the proxyfied requests (accepts: "warn" | "debug" | "info" | "error" | "silent")
 
 -   Create a file `.env.local` (copy it from `.env`) to customize environment variables so you can simply run `yarn start`.
```
tokland, line 60: Down here there are references to why-did-you-render and craco, is it still working with vite? if not, we can remove references in the readme.

package.json
```diff
@@ -10,29 +10,30 @@
         "url": "git+https://github.com/eyeseetea/metadata-synchronization.git"
     },
     "dependencies": {
```
tokland, line 12: I did yarn add -D depcheck and yarn depcheck, some dependencies are shown as unused. Not really worried about devDependencies, but the others we should check, to reduce the zip.

scripts/run.ts
```diff
@@ -136,7 +137,7 @@ function build(args: BuildArgs): void {
         const fileName = `${variant.file}.zip`;
         const manifestType = variant.type === "widget" ? "DASHBOARD_WIDGET" : "APP";
 
-        run(`react-scripts build && cp -r i18n icon.png build`);
+        run(`vite build && cp -r i18n icon.png build`);
```
tokland: Now that we are refactoring this, let's do it more generic, calling a "yarn ..." script.

scripts/run.ts
```diff
@@ -181,14 +182,14 @@ function startServer(args: StartServerArgs): void {
     }
 
     Object.assign(process.env, {
-        REACT_APP_PRESENTATION_TYPE: variant.type,
-        REACT_APP_PRESENTATION_VARIANT: variant.name,
-        REACT_APP_PRESENTATION_TITLE: variant.title,
+        VITE_PRESENTATION_TYPE: variant.type,
+        VITE_PRESENTATION_VARIANT: variant.name,
+        VITE_PRESENTATION_TITLE: variant.title,
```
tokland: We have two blocks with this, let's DRY it.

scripts/run.ts
```diff
@@ -181,14 +182,14 @@ function startServer(args: StartServerArgs): void {
...
         PORT: args.port,
     });
 
     run("yarn localize && d2-manifest package.json manifest.webapp");
-    run("react-scripts start");
+    run("vite");
```
tokland: same, let's use yarn scripts.

src/data/instance/InstanceD2ApiRepository.ts
```diff
@@ -214,13 +216,18 @@ export class InstanceD2ApiRepository implements InstanceRepository {
...
             url: data.type === "local" ? this.instance.url : data.url,
             version: data.type === "local" ? this.instance.version : data.version,
             username: data.type === "local" ? this.instance.username : data.username,
-            password: data.type === "local" ? this.instance.password : this.decryptPassword(data.password),
+            password:
+                data.type === "local"
+                    ? this.instance.password
+                    : data.password
+                    ? await this.decryptPassword(data.password)
+                    : "",
```
tokland: minor, but we try to avoid nested ? conditionals, proposal `data.type === "local" ? this.instance.password : await this.decryptPassword(data.password || "")`

src/data/metadata/MetadataD2ApiRepository.ts
```diff
@@ -102,7 +102,7 @@ export class MetadataD2ApiRepository implements MetadataRepository {
                 metadataPackage.dataStores = await d2ApiDataStore.getDataStores({ namespaces: ids });
             }
             const dataStoresMetadata = await this.getDataStoresMetadata(ids);
-            const responseWithDataStores = { ...metadataPackage, ...dataStoresMetadata } as T;
+            const responseWithDataStores = { ...metadataPackage, ...dataStoresMetadata } as MetadataPackage<T>;
```
tokland, line 106: This is not strictly related to the upgrade, but was this something that surfaced because of the TS upgrade? how it was working before?

src/domain/events/builders/EventsPayloadBuilder.ts
```diff
@@ -64,7 +64,7 @@ export class EventsPayloadBuilder {
 
         const directIndicators = programIndicators.map(({ id }) => id);
         const indicatorsByProgram = _.flatten(
-            programs?.map(({ programIndicators }: Partial<D2Program>) => programIndicators?.map(({ id }) => id) ?? [])
+            programs?.map(({ programIndicators }) => programIndicators?.map(({ id }) => id) ?? [])
```
tokland, line 67: According to the types, programs and programsIndicators cannot be undefined. But OK, let's keep them; this is probably not the right PR for such changes. It could simply be that the type changed (no problem), or that it's defensive code because at runtime it was undefined (problematic!).

src/domain/events/builders/EventsPayloadBuilder.ts
```diff
@@ -119,7 +119,7 @@ export class EventsPayloadBuilder {
         const onlyMetadataIds = metadataIds.filter(id => !DataStoreMetadata.isDataStoreId(id));
         const cleanIds = onlyMetadataIds.map(id => _.last(id.split("-")) ?? id);
         const metadataRepository = await this.getMetadataRepository(instance);
-        return metadataRepository.getMetadataByIds<T>(cleanIds, eventsFields);
+        return metadataRepository.getMetadataByIds<T>(cleanIds, eventsFields) as Promise<EventsMetadataExtracted>;
```
tokland, line 122: Not ideal, but again, this isn't the PR for that. The app's typing is generally lacking: we use blind type assertions that should be backed by purity-ts codec in runtime.

src/domain/events/builders/EventsPayloadBuilder.ts
```diff
@@ -147,3 +147,11 @@ export class EventsPayloadBuilder {
         }
     }
 }
+
+// TODO: remove this workaround for typing extractMetadata result
+// the underlying issue is in the getMetadataByIds typings
+type EventsMetadataExtracted = {
+    programs?: Pick<Program, "id" | "programType" | "programStages" | "programIndicators">[];
+    programIndicators?: { id: Id }[];
+    programStages?: { id: Id }[];
```
tokland: minor, Ref[]

src/index.tsx
```diff
@@ -17,7 +17,7 @@ declare global {
     }
 }
 
-const isDev = process.env.NODE_ENV === "development";
+const isDev = import.meta.env.MODE === "development";
```
tokland: A proposal. Again, now that we are touching this infrastructure, let's see how to do it more cleanly. What about having all the references to `import.meta.env` go to a single module (with get/set, and good types for the known variables)? This way we have all the Vite-related global variables in a single place. What do you think?

src/presentation/react/core/components/test-wrapper/utils.tsx
```diff
@@ -47,7 +47,9 @@ export function removeParentheses(string: string) {
 }
 
 export function isClassComponent(component: any) {
-    return typeof component === "function" && !!component.prototype.isReactComponent ? true : false;
+    return typeof component === "function" && !!component.prototype && !!component.prototype.isReactComponent
+        ? true
+        : false;
```
tokland: it was already there, but `conditions ? true : false` -> `condition`

src/presentation/utils/wdyr.ts
```diff
@@ -2,10 +2,12 @@
...
-    const whyDidYouRender = require("@welldone-software/why-did-you-render");
+if (import.meta.env.MODE === "development") {
+    import("@welldone-software/why-did-you-render").then(whyDidYouRenderModule => {
+        const whyDidYouRender = whyDidYouRenderModule.default;
 
-    whyDidYouRender(React, {
-        trackAllPureComponents: true,
+        whyDidYouRender(React, {
+            trackAllPureComponents: true,
+        });
```
tokland, line 12: oh, so why-did-you-render works with vite?

src/types/d2-api.ts
```diff
@@ -3,7 +3,39 @@ import { D2Api, D2ApiDefinition } from "@eyeseetea/d2-api/2.36";
...
+    Pager,
+    Stats,
+    DataValueSetsPostResponse,
+    SelectedPick,
+    DataStoreKeyMetadata,
+    D2Model,
+    Id,
+    Ref,
+    GetOptions,
+} from "@eyeseetea/d2-api/2.36";
```
tokland, line 38: so we are still using a d2-api with 2.36. I see the "evaluate impact" note. Probably that task should be a separate PR, I agree, it won't be trivial.

vite.config.ts
```diff
@@ -0,0 +1,96 @@
...
+                    "buffer",
+                    "util",
+                    "os",
+                    "events",
+                    "process",
+                    "path",
+                    "http",
+                    "https",
+                    "zlib",
+                    "vm",
```
tokland: That's a lot of polyfills. I guess you added the ones that appear as warnings during a yarn build? Okay, it is what it is; probably some of them are our fault, like having node-fetch in d2-api.

tokland (commented):
In-line comments:

src/data/metadata/MetadataD2ApiRepository.ts
MatiasArriola: Looks like type inference improved or some bug was fixed in TS. 
This function had an explicit return type `Promise<MetadataPackage<T>>`. Here we were casting `{ ...metadataPackage, ...dataStoresMetadata } as T` and then returning it, which is incorrect because `T` won't overlap `MetadataPackage<T>`

src/presentation/utils/wdyr.ts
MatiasArriola: Yes, although it was not working with the original changes. I just fixed it in `vite.config.ts` by passing `jsxImportSource` to the `react` plugin.

vite.config.ts
MatiasArriola: I remember adding them to solve warnings and runtime errors, but now I removed that `include` prop entirely and it works.

Also we get a warning during build: `node_modules/vm-browserify/index.js (110:11) Use of eval in "node_modules/vm-browserify/index.js" is strongly discouraged as it poses security risks and may cause issues with minification.`. I tried to exclude `vm` but looks like a transitive dependency and I get other warnings when doing so.

MatiasArriola:
Thanks for your review @tokland! 
The PR is updated with all changes, and the description now includes pending items that we might consider for creating new tasks.

README.md
```diff
@@ -56,11 +56,9 @@ Notes:
 
 -   The optional environment variable `VITE_DHIS2_AUTH=USERNAME:PASSWORD` forces some credentials to be used by the proxy. This variable is usually not set, so the app has the same user logged in at `VITE_DHIS2_BASE_URL`.
 
--   The optional environment variable `VITE_PROXY_LOG_LEVEL` can be helpful to debug the proxyfied requests (accepts: "warn" | "debug" | "info" | "error" | "silent")
-
 -   Create a file `.env.local` (copy it from `.env`) to customize environment variables so you can simply run `yarn start`.
 
--   [why-did-you-render](https://github.com/welldone-software/why-did-you-render) is installed, but it does not work when using standard react scripts (`yarn start`). Instead, use `yarn craco-start` to debug re-renders with WDYR. Note that hot reloading does not work out-of-the-box with [craco](https://github.com/gsoft-inc/craco).
+-   [why-did-you-render](https://github.com/welldone-software/why-did-you-render) is installed and runs in development mode by default.
```
tokland: Oh, nice, it's always active. Not sure if it slows down the app on development, but the tool is very useful.

Maybe we can add a reference to the file path to customize it (src/presentation/utils/wdyr.ts). Right now it only tracks pure components (the ones with React.memo), which seems a good default.

package.json
```diff
@@ -142,15 +125,12 @@
...
         "ts-mockito": "^2.6.1",
         "ts-node": "^10.8.2",
         "typescript": "4.9.5",
         "vite": "^4.0.0",
         "vite-plugin-node-polyfills": "^0.24.0",
         "vitest": "^3.2.4",
-        "wait-on": "6.0.0",
         "yargs": "15.4.1"
     },
     "resolutions": {
```
tokland, line 139: I get this output:

```bash
$ npx depcheck
Missing dependencies
* espree: ./.eslintrc.js
* dotenv: ./src/scheduler/cli.ts
* d2: ./src/presentation/react/core/components/module-list-table/ModuleListTable.tsx
* debug: ./src/data/migrations/tasks/01.instances-by-id.ts
* mochawesome-report-generator: ./cypress/support/runner.js
```

Can you check if they are really problematic? Also, what about adding depcheck as package and script, it's proven quite useful!

src/presentation/webapp/Root.tsx
```diff
@@ -29,7 +30,7 @@ import { About } from "../react/core/components/about/About";
 
 const Root: React.FC = () => {
     const { api, compositionRoot } = useAppContext();
-    const appVariant = getAppVariant();
+    const appVariant = config.appPresentationVariant ?? "core-app";
```
tokland: The rule of thumb is: don't defer default or fallback values to later code. If "core-app" is the default, let's make appPresentationVariant return that value directly so it's no longer a Maybe. This centralizes the fallback decision at a single initial point.

src/utils/Config.ts
```diff
@@ -0,0 +1,64 @@
+import { Maybe } from "../types/utils";
+
+export class Config {
```
tokland: Nice, now we have all the environment thingies in a single place.

src/utils/Config.ts
```diff
@@ -0,0 +1,64 @@
...
+    private static instance: Config;
+    private constructor() {}
+
+    public static getInstance(): Config {
+        if (!Config.instance) {
+            Config.instance = new Config();
+        }
+        return Config.instance;
```
tokland: I guess that was necessary because on every import the Config object was recreated. Maybe we can add a comment about why we need to cache it, as usually we would not do it.

Note that building a `new MyClass` is as expensive as its constructor. Here the constructor does nothing. So `new Config()` should be as cheap as it gets.

src/utils/Config.ts
```diff
@@ -0,0 +1,64 @@
...
+     */
+    get appPresentationVariant(): Maybe<AppVariant> {
+        const variant = this.presentationVariant;
+        return this.isValidAppVariant(variant) ? variant : undefined;
+    }
+
+    get presentationType(): Maybe<PresentationType> {
+        const type = import.meta.env.VITE_PRESENTATION_TYPE;
+        return type === "webapp" || type === "widget" ? type : undefined;
+    }
```
tokland, line 30: As discussed, let's try to remove all the Maybe that make sense to remove. And check in the usage that we don't have default values lingering.

vite.config.ts
tokland: This may be a problem, as we can get the whole app flagged as insecure (even if that particular use of eval on vm-browserify is not dangerous, nobody will check that). Review with your PM if this requires further attention, I think it might.

MatiasArriola: Well, sincerely I was tempted to leave it as it was because of time constrains, but after diving in this issue I realized that the `crypto.ts` util was pulling in all node `crypto` polyfills, while we are using the Web Crypto api for the browser environment! This in turn brought other deps such as `vm` with the "insecure eval" message.
`"@peculiar/webcrypto"` module should be loaded only from node environments, so I've added a shim for that resolution in vite. 
This change also helped save around 200kb of bundle size!

package.json
MatiasArriola: I've added depcheck as a devDependency and added it as script.
The only missing dep I fixed is `espree`, which is used in the eslint configuration for the cypress project.

All the others were already missing before this PR, and are included by other dependencies. Verified with `yarn why {package}`. If you are OK with this, we could try to fix them in the future.

Also, even though I've added depcheck, looks like it is deprecated now, and a good alternative mentioned in their readme is [knip](https://knip.dev/). I've added yet another pending note in this PR description to consider implementing it.

src/utils/Config.ts
MatiasArriola: This was not necessary at all. TBH copilot scaffolded this class for me and I thought it was a good idea to keep it with that singleton pattern. Also, as far as I know the import will be cached automatically so the config should be instantiated just once.

src/presentation/webapp/Root.tsx
MatiasArriola: That's a good improvement. I was worried to change the original behavior, but now I've updated `appPresentationVariant` and `presentationType` and I think it is still safe.
I left the `Maybe` in other two cases: `presentationTitle` and `presentationVariant`. There is one `presentationVariant` usage I don't fully understand in `src/presentation/widget/pages/Root.tsx`, so for safety I kept as it is.

package.json
tokland: Yes, just add a note to the PR description about this future task.

tokland (approved):

MiquelAdell:
Hi, I'm just testing this but a reminder. Changing from REACT to VITE means that we need to change 
REACT_APP_DHIS2_BASE_URL
REACT_APP_DHIS2_AUTH
to
VITE_DHIS2_BASE_URL
VITE_DHIS2_AUTH

in the .env.local

That would be something ideal for "notes to the tester" :-P

PR #1015, opened 2025-06-24
[feature]: Add line lists to eventVisualizations payload

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8699f0vdj

### :memo: Implementation
- [x] Fetch event visualizations from DHIS2 for dashboards and add it to the export payload

https://github.com/user-attachments/assets/50a53a0e-67fe-4e0e-af0f-6416af9193ee

### :bookmark_tabs: Others

#8699f0vdj

### DIFF
--- a/src/domain/metadata/builders/MetadataPayloadBuilder.ts
+++ b/src/domain/metadata/builders/MetadataPayloadBuilder.ts
@@ -378,7 +378,7 @@ export class MetadataPayloadBuilder {
             eventVisualizations: lineListVisualizations,
         };
     }
-
+    
     private async addRowsToVisualizations(
         originInstance: Instance,
         visualizations: Visualization[]


src/data/metadata/EventVisualizationD2ApiRepository.ts
```diff
@@ -0,0 +1,25 @@
...
+export class EventVisualizationD2ApiRepository implements EventVisualizationRepository {
+    private api: D2Api;
+
+    constructor(instance: Instance) {
+        this.api = getD2APiFromInstance(instance);
+    }
+
+    async getByIds(ids: Id[]): Promise<EventVisualization[]> {
+        const { eventVisualizations } = await this.api
+            .get<{ eventVisualizations: EventVisualization[] }>("/eventVisualizations", {
```
tokland: Can we add a comment explaining why are we doing a manual get request instead of using `api.metadata.get`? this way, when d2-api is updated, it will be easier to spot.

src/domain/metadata/builders/MetadataPayloadBuilder.ts
```diff
@@ -333,4 +345,24 @@ export class MetadataPayloadBuilder {
...
+        dashboard: Dashboard,
+        originInstanceId: string
+    ): Promise<{ eventVisualizations: EventVisualization[] }> {
+        const defaultInstance = await this.getOriginInstance(originInstanceId);
+        const eventVisualizationsRepository = this.repositoryFactory.eventVisualizationRepository(defaultInstance);
+
+        const eventVisualizationIds = _(dashboard.dashboardItems)
+            .map(dashboardItem => dashboardItem.eventVisualization?.id)
+            .compact()
+            .value();
```
tokland, line 368: It's unlikely we'll have repeated event visualizations, but as a general rule, we `.uniq()` a list of ids to get. (that's compatible with the repository method also doing it)

src/domain/metadata/builders/MetadataPayloadBuilder.ts
```diff
@@ -333,4 +345,24 @@ export class MetadataPayloadBuilder {
...
+        const eventVisualizations = await eventVisualizationsRepository.getByIds(eventVisualizationIds);
+
+        return {
+            eventVisualizations: eventVisualizations.filter(
+                eventVisualization => eventVisualization.type !== "LINE_LISTING"
```
tokland: The function is `IncludeLineListings`, but we have `filter(type !== "LINE_LISTING")`, so we are in fact excluding them.

src/domain/metadata/repositories/EventVisualizationRepository.ts
```diff
@@ -0,0 +1,11 @@
+import { Id } from "../../common/entities/Schemas";
+import { Instance } from "../../instance/entities/Instance";
+import { EventVisualization } from "../entities/MetadataEntities";
+
+export interface EventVisualizationRepositoryConstructor {
```
tokland: Is this interface used anywhere?

src/domain/metadata/repositories/EventVisualizationRepository.ts
```diff
@@ -0,0 +1,11 @@
...
+    new (instance: Instance): EventVisualizationRepository;
+}
+
+export interface EventVisualizationRepository {
```
tokland:
```
$ ls src/domain/metadata/repositories/
EventVisualizationRepository.ts  MetadataRepository.ts
```

We’ve been using a single MetadataRepository for all other models, so it feels odd that we suddenly need a dedicated repository for this one. Is that something you worked on with @xurxodev?

src/data/metadata/EventVisualizationD2ApiRepository.ts
```diff
@@ -0,0 +1,25 @@
...
+                fields: ":all",
+                filter: `id:in:[${ids.join(",")}]`,
```
tokland: We can add a _.uniq here so we allow the caller to forget about it.

src/domain/metadata/builders/MetadataPayloadBuilder.ts
deeonwuli: you are right. we should instead include line lists with `filter(type === "LINE_LIST"`

src/domain/metadata/repositories/EventVisualizationRepository.ts
deeonwuli: no. fixed

src/domain/metadata/repositories/EventVisualizationRepository.ts
xurxodev: I think in order of priority:
1 - If it's possible to use existed functions in metadata repository I think we should use it
2 - If it's not possible to use existed functions we should think if it makes sense to add it to metadata repository
3 - if It's a very specific functionality with different request to the API maybe makes sense a different repository

Reviewing the repository, it has only getById with :all fields.

I think that we can to use MetadataRepository.  In getMetadataByIds is possible to pass fields as argument. 
Maybe necessary add EventVisualization entity to MetadataEntities to work

deeonwuli: noted 📝

src/domain/metadata/builders/MetadataPayloadBuilder.ts
```diff
@@ -333,4 +345,28 @@ export class MetadataPayloadBuilder {
         });
         return { ...program, programRules };
     }
+
+    private async requestAndIncludeLineListings(
```
tokland, line 358: The code itself is fine and retrieves the lineListings of the eventVisualizations as intended. However, it’s not clear why, when including dashboards, we are only interested in that particular subset of visualizations. We should add a comment here to clarify the reasoning.

MiquelAdell:
Hi @deeonwuli when you address the comments from Arnau please also fix the conflicts so we can merge the PR.

Thanks!

deeonwuli:
> Hi @deeonwuli when you address the comments from Arnau please also fix the conflicts so we can merge the PR.
> 
> Thanks!

All done @MiquelAdell

PR #1021, opened 2025-07-29
[fix] MetadataPayloadBuilder caching strategy

MatiasArriola:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8699k06vd #8699k06vd

### :memo: Implementation

Consider the current builder parameters when using cached values. Some references might be missing in some requests but not in others, for the same id, leading to missing dependencies in the final package.

This ocurred for a userGroup and its users:
- The userGroup was first resolved as a transitive dependency of a program, without its users. 
- The same userGroup was selected in the metadata sync list. Its users were expected, but they were missing due to the earlier cached response.

Other: 
- Added a test case to `/MetadataPayloadBuilder.spec.tsx` to reproduce this bug.
- Added a debug method to `MetadataPayloadBuilder` (off by default). It was helpful for me to debug the dependency resolution, so I thought we could keep it in case we need it in the future.


Ramon-Jimenez (approved):
Working great, thanks @MatiasArriola

PR #1027, opened 2025-09-02
Fix period order visualizations

anagperal:
### :pushpin: References

-   **Issue:** Closes #869a9zgvp [Metadata Sync is not getting all data on visualization rows](https://app.clickup.com/t/869a9zgvp)

### :memo: Implementation
- [x] Add VisualizationsRepository (using metadata endpoint forcing fields to get rows returns rows empty always)
- [x] Get visualizations rows from VisualizationsRepository and add to visualizations in MetadataPayloadBuilder

BEFORE:

[WITHOUT_FIX.json](https://github.com/user-attachments/files/22092067/WITHOUT_FIX.json)

AFTER:

[FIXED.webm](https://github.com/user-attachments/assets/fe546f21-603c-49b4-91a7-e3bef7314046)

[FIXED.json](https://github.com/user-attachments/files/22092066/FIXED.json)

[dev_eyeseetea_payload_only_visualizations.json](https://github.com/user-attachments/files/22097123/dev_eyeseetea_payload_only_visualizations.json)

[dev_eyeseetea_payload.json](https://github.com/user-attachments/files/22097124/dev_eyeseetea_payload.json)

Tested from https://play.im.dhis2.org/stable-2-40-8-2 to https://play.im.dhis2.org/stable-2-41-5

### :bookmark_tabs: Others


ifoche:
Task linked: [CU-869a9zgvp Metadata Sync is not getting all data on visualization rows](https://app.clickup.com/t/869a9zgvp)

src/data/visualizations/VisualizationsD2Repository.ts
```diff
@@ -0,0 +1,37 @@
...
+import { Visualization } from "../../domain/visualizations/entities/Visualization";
+import {
+    GetVisualizationsOptions,
+    VisualizationsRepository,
+} from "../../domain/visualizations/repositories/VisualizationsRepository";
+import { D2Api } from "../../types/d2-api";
+import { getD2APiFromInstance } from "../../utils/d2-utils";
+import { apiToFuture } from "../common/utils/api-futures";
+
+export class VisualizationsD2Repository implements VisualizationsRepository {
```
tokland: [minor] Jorge recommends using the singular name for a repository: "VisualizationRepository". But check the rest of the project and follow the existing practice.

src/data/visualizations/VisualizationsD2Repository.ts
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    async getByIds(ids: Id[], options: GetVisualizationsOptions): Promise<Visualization[]> {
+        const fields = options.onlyRows ? "id,rows[dimension,items[id,name]]" : "*";
+
+        const $reqVisualizations = ids.map(id => {
+            return apiToFuture(
+                this.api.get<Visualization>(`/visualizations/${id}`, {
+                    fields: fields,
+                })
```
tokland: Whenever we need to do a manual request, we should specify with a comment why we cannot use the fully typed endpoints. 

Note that newer version of d2-api provide `api.models.MODEL.getById(id)`, but probably not the version used here.

src/data/visualizations/VisualizationsD2Repository.ts
```diff
@@ -0,0 +1,37 @@
...
+            );
+        });
+
+        return Future.parallel([...$reqVisualizations], {
```
tokland: why the `[...array]` ?

src/data/visualizations/VisualizationsD2Repository.ts
```diff
@@ -0,0 +1,37 @@
...
+            concurrency: 5,
+        })
+            .map(visualizationsResponse => visualizationsResponse)
```
tokland: The map is a no-op, can be removed.

src/data/visualizations/VisualizationsD2Repository.ts
```diff
@@ -0,0 +1,37 @@
...
+            .toPromise();
```
tokland: So we are using Future for the parallel, right? I see that strategy is not typical in the app, it may be overkill. Alternatives: 1) do a normal sequential promises request if we can do without the parallel. 2) Use a pool of request (some library in npmjs for promises?) if we really need the parallel feature. 3) keep as-is.

src/domain/metadata/builders/MetadataPayloadBuilder.ts
```diff
@@ -115,9 +121,14 @@ export class MetadataPayloadBuilder {
             categoryCombos,
             categoryOptions,
             categoryOptionCombos,
+            visualizations,
             ...rest
         } = metadataWithoutDuplicates;
 
+        const visualizationsWithRows = visualizations
+            ? await this.addRowsToVisualizations(originInstance, visualizations as Visualization[])
```
tokland, line 129: [nothing to do] It's unfortunate that we need this type-unsafe assertion, but I see that's how it's done in all the app

src/domain/metadata/builders/MetadataPayloadBuilder.ts
```diff
@@ -131,6 +142,9 @@ export class MetadataPayloadBuilder {
             ...(categoryOptionCombos && {
                 categoryOptionCombos: this.excludeDefaultMetadataObjects(categoryOptionCombos, removeCategoryObjects),
             }),
+            ...(visualizations && {
+                visualizations: visualizationsWithRows,
```
tokland: it's conceptually more fitting that the checked var and the value is the same, so `visualizationsWithRows`. So we set to `undefined`, not `[]` above in the code for the no-value.

src/domain/metadata/builders/MetadataPayloadBuilder.ts
```diff
@@ -317,4 +331,21 @@ export class MetadataPayloadBuilder {
         });
         return { ...program, programRules };
     }
+
+    private async addRowsToVisualizations(
+        originInstance: Instance,
+        visualizations: Visualization[]
+    ): Promise<Visualization[]> {
+        const visualizationsRepository = await this.repositoryFactory.visualizationsRepository(originInstance);
+        const visualizationIds = visualizations?.map(visualization => visualization.id) || [];
```
tokland: The type is `Visualization[]`, do we need the `?` ?

src/data/visualizations/VisualizationsD2Repository.ts
```diff
@@ -0,0 +1,37 @@
...
+    constructor(instance: Instance) {
+        this.api = getD2APiFromInstance(instance);
+    }
+
+    async getByIds(ids: Id[], options: GetVisualizationsOptions): Promise<Visualization[]> {
+        const fields = options.onlyRows ? "id,rows[dimension,items[id,name]]" : "*";
+
+        const $reqVisualizations = ids.map(id => {
+            return apiToFuture(
+                this.api.get<Visualization>(`/visualizations/${id}`, {
```
tokland: That looks off. The option determines if we return all the fields or only id/rows. But the output type is Visualization, where all fields are available.

Not sure how will that fit in the current code of mdsync, but the general approach is to have two different methods. Here it's onlyRows: false used? if not, let's skip it, with the general idea of not implementing "future code" (the other option would be conditional type, not recommended for our own code).

In summary, the specified return type of the method should be the real returned type.

tokland (commented):
In-line comments:

src/data/visualization/VisualizationD2Repository.ts
```diff
@@ -0,0 +1,64 @@
...
+            })
+            .getData();
+
+        return d2Visualizations.objects.map(d2Visualization => this.mapD2VisualizationToVisualization(d2Visualization));
+    }
+
+    private mapD2VisualizationToVisualization(d2Visualization: D2Visualization): Visualization {
+        return {
+            id: d2Visualization.id,
+            rows: (d2Visualization.rows as D2VisualizationRow[]).map(row => ({
```
tokland, line 32: We need a comment here about needing this type assertion. I see that we have `rows: unknown[];  series: unknown[];`. This is something that will need fixing in d2-api generate-schemas. But for now, let's have the comment.

src/domain/metadata/builders/MetadataPayloadBuilder.ts
```diff
@@ -335,16 +335,16 @@ export class MetadataPayloadBuilder {
     private async addRowsToVisualizations(
         originInstance: Instance,
         visualizations: Visualization[]
-    ): Promise<Visualization[]> {
+    ): Promise<Visualization[] | undefined> {
         const visualizationsRepository = await this.repositoryFactory.visualizationsRepository(originInstance);
-        const visualizationIds = visualizations?.map(visualization => visualization.id) || [];
+        const visualizationIds = visualizations.map(visualization => visualization.id) || [];
```
tokland: Can we remove `|| []` ?

src/domain/metadata/builders/MetadataPayloadBuilder.ts
```diff
@@ -335,16 +335,16 @@ export class MetadataPayloadBuilder {
...
 
-        const visualizationsWithOnlyRows = await visualizationsRepository.getByIds(visualizationIds, {
-            onlyRows: true,
-        });
+        if (visualizationIds.length === 0) return undefined;
```
tokland: Instead of this check here, I'd move it to the repo method getByIds.

src/domain/visualization/entities/Visualization.ts
```diff
@@ -0,0 +1,17 @@
...
+    name: string;
+};
+
+type VisualizationRow = {
+    id: string;
+    dimension: string;
+    items: VisualizationRowItem[];
+};
+
+export type Visualization = {
```
tokland: minor: top-down types is usually more readable, the "important" type is the first you see when you open the file.

tokland (approved):
All good!

PR #1028, opened 2025-09-03
fix: sync tei enrollment attributes depending on dhis target version

MatiasArriola:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869aakwnk #869aakwnk

### :memo: Implementation

- Events Sync was failing for DHIS 2.40 because the server expects `attributes` to be included inside `trackedInstanceEntity.enrollments`
- For versions >= 2.41 `attributes` should be included at the `trackedInstanceEntity` level only

**Solution:** 
-  Default to fetching and posting attributes at both the `trackedInstanceEntity` and `trackedInstanceEntity.enrollments` levels
- For versions >= 41, remove `attributes` from enrollments (override with empty array), using `PackageTransformations`
- Added a test case for this scenario

**Extra:**
- While testing syncing events migration to a v41+ target instance, I discovered a bug where multiple unnecessary requests were being made to fetch events. 

```js
const pageCount = Math.ceil(firstResponse.total || 0 / firstResponse.pageSize);
```

Was causing to incorrectly fetch as many pages as the `total` field included. There is a subtle operator precedence bug here. Found another similar case but was correct. Fixed it as part of this PR and took the opportunity to refactor for code reuse. Also now it should avoid an unnecessary second page request if `pageCount` is 1 or 0.

- Manually tested syncing events with teis from v40 to v41 using sierra db and behaves as expected


xurxodev (changes requested):
thanks @MatiasArriola 

First I've done a code review directly in the PR.

Great that you have discovered a bug in the pagination and now it's fixed and with tests.

Reviewing the code I just have one change to request.

1 - The payload transformations is a dhis2 data layer concern, not domain. For this reason, the responsible to invoke the transformations should be the repository like in save method in metadataD2ApiRepository.

I know that some use cases about packages are invoking transformations, but it's a TODO improvement in this project.

TODO: Transformations should be realized in data layer and transformationRepository should not be named so because only should be invoked by repositories and not use cases or similar.

Step by step we will refactor this in the future, but for the new code we should try to follow the clean architecture rules.

MatiasArriola:
Thanks @xurxodev! Good observation, I agree this is a data concern. Updated the PR to call the transformation from inside the `TEID2ApiRepository`.

As you guessed I went this way for consistency, because transformationRepo was mostly used from useCases. Also I thought about one repository calling another repository is not correct from clean arch perspective... But now that you mention it, `TransformationRepository`, in fact is not a Repository and should be moved and renamed in future refactors. 

One caveat: at the repository level I don't have available the originInstance version, so I've omitted the `origin?: number` parameter from the `mapPackageTo` call. It will fall back to the `API_VERSION` constant which is set to `36` now, which will work in our case, but we might need to think about it during the future refactor.

Ramon-Jimenez (approved):
Working great @MatiasArriola , thanks. Waiting on @xurxodev 's feedback before merging this one.

xurxodev (changes requested):
thanks @MatiasArriola 

Now the transformations are done in the data layer.

I have some change to request in this case in the tests.

The tests in this project for transformations are under transformation folder. The integration tests don't validate transformations. And there are a separate integration tests for every repository type like events, metadata, aggregated

To follow the same approach ...

Can you remove the TEI test about transformations from instegration sync-events.spec.ts and create transformations tests for TEI under transformation folder?

In the future we should create sync-tei.spec.ts integration tests, but I think that is not responsibility of this PR

MatiasArriola:
Latest changes:
- Added unit tests for the transformation in `src/data/transformations/__tests__/PackageTransformations/teiTransformations.spec.ts`
- Reverted the integration test from `src/data/metadata/__tests__/integration/sync-events.spec.ts`
- Added integration tests for the transformation in `src/data/transformations/__tests__/integration/teiTransformations-api.spec.ts`

I’m not fully happy with the new integration test: it has a lot of boilerplate and can be hard to reason about. That said, I think it’s valuable to document and verify the new behavior. If you still think it should be removed, we can do it.

Originally, I added the test in `sync-events` to reuse the existing setup. Even though the filename says "events", it’s actually testing the `EventsSyncUseCase`, which also covers `trackedEntityInstances` and `dataValues`.

Regarding the existing transformation tests:
- `src/data/transformations/__tests__/AggregatedTransformationD2Api.spec.ts`
- `src/data/transformations/__tests__/EventsTransformationD2Api.spec.ts`
- `src/data/transformations/__tests__/MetadataTransformationD2Api.spec.ts`

These are almost the same test with different names/twists. Ideally, there should just be a single `TransformationD2ApiRepository.spec.ts`. None of the existing `PackageTransformations` had unit tests before.

There was `src/data/transformations/__tests__/integration/transformations-api-38.spec.ts` but only for metadata. It is in fact testing the same surface as `src/data/metadata/__tests__/integration/sync-metadata.spec.ts`, only that including a case specific for the users transformation.

Note that all tests under `src/data/metadata/__tests__/integration` are integration tests of UseCases and PayloadBuilders.

xurxodev (approved):
Thanks @MatiasArriola

Now, the new test follows the same approach as the existing tests.

The current integration tests are in the data layer and here TEI and Events are separate concepts. Propably, data layer is not the best place for integration tests using the use use cases

It’s possible that the existing tests need some refactoring and improvement.

Thinking about future developers joining the project, I think the best option is to address this refactor as a whole in the same PR, instead of creating a different approach only for TEI in this PR, since future contributors would have to figure out which one to follow.

Probably @Ramon-Jimenez  or @adrianq  should add a task in the metadata sync backlog to evaluate the refactor of the tests.

I've tested this PR syncing against a 2.40 and 2.41 destination sever and the payload is built as expected and the syncs are successful

PR #1034, opened 2025-10-21
Feature/user default inclusion modes

gqcorneby:
### :pushpin: References

-   **Issue:** Closes #869aut7cn [metadatasync  user-usergrups-userroles by default](https://app.clickup.com/t/869aut7cn)

### :memo: Implementation
- add UserSettings that is saved consistently in user data store (regardless of selected storage)
- set user defaults in settings page
- use user default inclusion modes when creating on demand

https://github.com/user-attachments/assets/843460e9-25a5-4ca0-9b29-39e5c4cba632

### :bookmark_tabs: Others


gqcorneby:
Workflow failures are fixed here: https://github.com/EyeSeeTea/metadata-synchronization/pull/1035

MiquelAdell:
Once the defaults are set I can not change them. I tried that in dev.est and in a play instance with a fresh install:

https://github.com/user-attachments/assets/613b567b-209f-4d4a-aad6-4ad760287a6b

gqcorneby:
Hi @MiquelAdell! Sorry about that! Should be fixed now

MiquelAdell (approved):

xurxodev (changes requested):
thanks @gqcorneby 

The exclude/include dependencies configuration is in ondemand and stored sync rules.

The change has affected to ondemand sync rules, but not to stored sync rules.

If we have default settings for this configuration, I think should work in both. 

To execute an ondemand sync rule and to create a new sync rule to save.

Here the video

https://github.com/user-attachments/assets/9f4b5d72-1b26-4fcb-978b-eb49da387d4d

gqcorneby:
Hi @xurxodev! Thanks for the feedback. I rechecked my earlier conversation with Miquel, and from what I understood, the defaults were meant to apply only to the three drop-downs in the manual (on-demand) sync modal.

Your suggestion to have them apply in the wizard as well makes sense. I’ll confirm the scope with @miquel before making any changes.

MiquelAdell:
Oh, I'm not sure what I defined as the scope but it would be great if the stored preferences would be applied to both places yes.
If you can @gqcorneby go ahead with it

thanks!

gqcorneby:
Hi @MiquelAdell and @xurxodev! I've applied the user defaults to the sync rule wizard. 

image

xurxodev (approved):
Thanks @gqcorneby

I approve the PR.

Minor comment for future consideration:

The inclusion models in userSettings and SyncRule are structured differently. 

We should avoid coupling these models. Currently, setDefaultInclusions and createOnDemand handle the mapping from UserSettingsInclusionsConfig to the sync rule's inclusion model, which creates coupling. 

This mapping should be handled outside the sync rule, as it is not its responsibility.

ifoche:
@MiquelAdell provided that the PRs approved by @xurxodev . Could you please proceed merging?

Thank you!

PR #1036, opened 2025-10-21
Fix start-scheduler and build-scheduler

MatiasArriola:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869awujfj

### :memo: Implementation

`start-scheduler` and `build-scheduler` stopped working as part of the changes in #1014 

Problem: the project started using ES modules features like `import.meta.env`, while the scheduler script was using CJS.

- For `start-scheduler`: replaced `ts-node` with `tsx`. With this change we can run the scheduler without no further changes, its faster and preferred in latest skeleton versions.
- For `build-scheduler`:
  - I tried to move to a vite build with ES modules instead of ncc CJS, or adding `"type": "module"` to `package.json` to have ncc generate a ES modules build, but couldn't get it working in a reasonable time.
  - Instead, made `Config` use `process.env` instead of `import.meta.env`. Changed `vite.config.js` to map the `process.env` values for the web build as well. This allows to use the same `Config` class for both vite web builds and ncc builds.
- Extra: fixed a type export missing in d2-api


MiquelAdell (approved):

vite.config.ts
```diff
@@ -27,6 +27,11 @@ export default defineConfig(({ mode }) => {
         define: {
             // Preserve process.env access for compatibility
             "process.env.NODE_ENV": JSON.stringify(env.NODE_ENV || mode),
+            // preserve process.env.VITE_ variables for backwards compatibility with CJS
+            "process.env.VITE_PRESENTATION_VARIANT": JSON.stringify(env.VITE_PRESENTATION_VARIANT),
+            "process.env.VITE_PRESENTATION_TITLE": JSON.stringify(env.VITE_PRESENTATION_TITLE),
+            "process.env.VITE_PRESENTATION_TYPE": JSON.stringify(env.VITE_PRESENTATION_TYPE),
+            "process.env.VITE_CYPRESS": JSON.stringify(env.VITE_CYPRESS),
```
tokland, line 34: This also caused issues for us in CPR. Just as a note, what we did there was use `vite-node` to run the scripts so that the `import.meta` references keep working.

That’s not to say this approach is wrong, moving everything to environment variables is reasonable.

tokland (approved):

PR #1041, opened 2025-11-13
Include TEIs referenced by events when using allTEIs option

MatiasArriola:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869b5ggyy #869b5ggyy

### :memo: Implementation

In Sync Event rules, when `allTEIs` is selected, but the period filter leaves out some TEI referenced by any event from that period, the sync was failing because of the missing TEI dependency.

- Added optional attr `trackedEntity` to the `ProgramEvent` type (it was already being mapped)
- Include in the Events Sync Payload all the TEIs referenced by the events if the "Sync all TEIs" option is used


Ramon-Jimenez (approved):
All good

PR #1043, opened 2025-12-11
Only show instances with permission

xurxodev:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869bd6ykk #869bd6ykk

### :memo: Implementation

- [x] Only show instances with permission
- [x] Add unit tests to validate all scenarios

#### Permission-based Filtering Scenarios

-   **Global Admin Access**: Global admin sees all instances, including those without explicit permissions
-   **Owner Access**: User sees instances where they are the owner (user.id matches instance.user.id)
-   **User Access Permissions**: User sees instances with direct user access permissions (userAccesses contains user with read permission)
-   **User Group Access Permissions**: User sees instances with group access permissions (userGroupAccesses contains a group the user belongs to with read permission)
-   **Public Read Access**: User sees instances with public read access (publicAccess starts with "r")
-   **No Permissions**: User with no permissions sees only the LOCAL instance
-   **Non-App Configurator**: User who is not an app configurator sees only LOCAL instance, even if they have other permissions (owner, public access, etc.)

#### Combined Scenarios

-   **Multiple Permission Types**: User with multiple permission types (owner, user access, group access, public access) sees all instances they have access to through any permission type

#### Expected Behavior

-   **LOCAL Instance Always Included**: The LOCAL instance is always included in the results, regardless of permissions
-   **Permission Filtering Works**: Instances without valid permissions are excluded from results
-   **App Configurator Requirement**: Only app configurators can see instances (except LOCAL), even if they have other permissions

#### Key Test Data Points

-   **User properties**: `isGlobalAdmin`, `isAppConfigurator`, `userGroups`
-   **Instance properties**: `user` (owner), `publicAccess`, `userAccesses`, `userGroupAccesses`
-   **Permission format**: Read permission is indicated by "r" in the first two characters of the access string (e.g., "r-------")

### :bookmark_tabs: Others


xurxodev:
@adrianq Ready the conflicts

PR #1046, opened 2026-02-12
Datastore metadata sync does not reflect selected namespaces and keys correctly

anagperal:
### :pushpin: References

-   **Issue:** Closes #869bvkyum [Datastore metadata sync does not reflect selected namespaces and keys correctly](https://app.clickup.com/t/869bvkyum)

### :memo: Implementation

- not sync excluded datastore keys
- download json with datastore data to be sync
- improve message in sync summary

[test.webm](https://github.com/user-attachments/assets/3488a3eb-fea5-4008-b8ea-398f934c6a0d)

### :bookmark_tabs: Others


ifoche:
Task linked: [CU-869bvkyum Datastore metadata sync does not reflect selected namespaces and keys correctly](https://app.clickup.com/t/869bvkyum)

xurxodev (changes requested):
thanks @anagperal

This PR addresses three real datastore sync bugs:
(a) deselecting a key after selecting its namespace no longer leaks the key into sync,
(b) the downloaded JSON is no longer empty because datastore content is now built on the download path and dumped to its own file, and
(c) the wizard summary reflects partial namespace selections accurately. The fixes attack the actual root causes (selection state propagation, missing builder call on the download path, summary derivation), not symptoms. Overall the direction is right; main concerns are around testability, a counter/UI mismatch and inconsistent error handling.

## 1. Should fix

- **`MetadataTable.tsx:617-690` — extract the datastore branch into a helper.** `handleTableChange` is now ~95 lines mixing two selection algorithms with table state and filter updates. The new datastore logic is the most fragile part of the PR and lives inline in a UI handler, which makes it both hard to read and hard to test. Extract something like `computeDataStoreSelection({ included, newlySelectedIds, newlyUnselectedIds, excludedIds, rows, parseChildren }) => { included, excluded }` and keep `handleTableChange` as a thin orchestrator. This also unlocks the unit tests below.

- **No regression tests.** Three real bugs fixed, zero tests added. The project already uses Vitest and has plenty of unit tests under `src/**/__tests__/*.spec.ts` (entities, mappers, utils, validators), so the testing infrastructure is in place — these would be plain unit tests, not e2e. Minimum to add:
  - **`DataStoreMetadata.buildFromKeys`** in `src/domain/data-store/__tests__/DataStoreMetadata.spec.ts` — covers the new `excludedDataStoreIds` filtering, the stricter `match.length !== 2` validation, and empty input. Pure static method, trivial to test.
  - **`isEmptyContent`** in `src/domain/synchronization/__tests__/utils.spec.ts` — `null`/`undefined`, primitives, `{}`, `{a: []}`, `{a: undefined}`, `{a: [1]}`, `{a: {}}`, `{a: {b: 1}}`. Pure function.
  - **The datastore selection algorithm**, *once extracted to a helper as per Should fix 
 point 1*. If it stays inline in `handleTableChange` it can't be unit-tested without React Testing Library (which the repo doesn't currently use). After extraction it becomes a pure function `computeDataStoreSelection({ included, newlySelectedIds, newlyUnselectedIds, excludedIds, rows, parseChildren }) => { included, excluded }` and the cases from the ticket map directly to test cases: (1) select namespace → all keys included, (2) after (1), deselect one key → namespace removed from `included`, other keys remain (this is the actual bug from the ticket), (3) deselect namespace → namespace and keys removed, (4) select a single key without touching the namespace → only that key. This is what would have caught the three bugs in CI before the fix.

- **`SummaryStep.tsx:611-621` — hard-coded `"[NS]"` literal duplicates `DataStoreMetadata.NS_SEPARATOR`.** The constant exists and `DataStoreMetadata.isNamespaceOnlySelected` already encapsulates the same check. Use them; otherwise the magic string is duplicated and silently breaks if the separator ever changes.

- **`SummaryStep.tsx:627` — counter doesn't match what's rendered.** The label says `DataStore [${dataStoreInfo.length}]` but the rendered list is `summaryInfo` (which is `dataStoreInfo` deduplicated by namespace). The number shown to the user can be larger than the items below it. Use `summaryInfo.length` so the count matches the visible list.

- **`DownloadPayloadFromSyncRuleUseCase.ts:249-251` — inconsistent error handling.** `mapDatastoreToDownloadItems` throws on a single-instance failure, which aborts the entire download (there's no try/catch upstream around the call). Compare with `mapToDownloadItems` for normal metadata, which catches per instance and returns an i18n error string that gets surfaced via the `errors` array. The datastore variant should follow the same pattern: catch per instance, return a user-friendly error string, accumulate into `errors`. Also `${error}` interpolates as `[object Object]` for non-Error throwables — use `error instanceof Error ? error.message : String(error)`.

- **i18n for new user-facing strings.** `"Namespace: X - Key: Y"` and `"All Keys"` are not wrapped in `i18n.t(...)`, while the other labels in `SummaryStep.tsx` are. Wrap them for consistency with the rest of the file.

## 2. Recommendations non blocking

- **Comments-as-naming in the new selection branch.** The `// Downward rule:` / `// Upward rule:` blocks describe _what_ the code does, not _why_. Once the helper is extracted (Should fix point 1), the comments can be replaced by descriptive function/variable names (`propagateNamespaceSelection`, `recomputeNamespaceParents`, etc.).

PR #1048, opened 2026-02-24
[fix]:  Preserve metadata selections when using "Select All" in metadata table

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869c32tz2

### :memo: Implementation
When "Select all" is clicked, the selection now merges IDs by metadata type, using `row.model.getMetadataType()`. The update in `handleTableChange` now:
- Identifies the current metadata type.
- Merges new selections with previous selections for that type.
- Keeps selections from other types unchanged.
- Calls `notifyNewSelection` with the merged selection and updated exclusions.
This ensures that bulk selection actions do not unintentionally clear selections for another metadata type

https://github.com/user-attachments/assets/a717e488-cf8e-44ba-a7df-9675fa1bb9c6

### :bookmark_tabs: Others

#869c32tz2


ifoche:
Task linked: [CU-869c32tz2 selecting "from all pages" excludes other metadata types](https://app.clickup.com/t/869c32tz2)

gqcorneby (commented):
Just noting it here, @deeonwuli and I had a discussion about some things regarding implementation that was easier communicated in a call 😁 

Other than the discussed changes, @MiquelAdell, there are some UI concerns that I noticed and wanted to see if it's something that we want to address in this PR

1. "select all" notification
This notification is no longer disappearing because of the updates here. It doesn't have to do with how the selection was done but more about how d2-ui checks if this notification should be shown. Currently it's checking if all items are selected by checking if `selection.length === total`. In this case, it's not equal because selection is 222 and length is 221. A workaround is to check if all is selected and then override the `selectionMessages`. 
Image

2. clear all
This is not related to this task at all but this action is only clearing selection for the current metadata type and not the others selected. So clicking clear isn't doing anything if constants is selected and I have a selected item in categories. 
Image

MiquelAdell:
> Other than the discussed changes, @MiquelAdell, there are some UI concerns that I noticed and wanted to see if it's something that we want to address in this PR
> 
> 1. "select all" notification
>   …
> 2. clear all
> …

both can be in another PR so we can close and merge this one. I'll create tasks for them
thanks!

xurxodev (approved):
thanks @deeonwuli

PR #1050, opened 2026-03-04
fix(packages): await metadataPayloadBuilder.build() in CreatePackageUseCase

ifoche:
## Summary

- Fix missing `await` on `metadataPayloadBuilder.build()` in `CreatePackageUseCase`, which caused a raw Promise to be stored as package contents instead of the resolved metadata payload — making all module-generated packages appear empty
- Remove unnecessary generic type parameters from `TransformationRepository.mapPackageTo/mapPackageFrom` to prevent Promise objects from being silently accepted as payload (compile-time safety)
- Add unit tests for `CreatePackageUseCase` (5 test cases)

## Related Tasks
- [Module package imports successfully but contains no data](https://app.clickup.com/t/869bvjxj9)

## Test plan
- [ ] Run `yarn test` — all 393 tests pass (48 files)
- [ ] Manual: create a module with a dataset, generate a package, verify the package contains the dataset metadata
- [ ] Manual: import the generated package on another instance, verify elements appear in the result summary
- [ ] Manual: download package as JSON, verify metadata objects are present
- [ ] Manual: compare package with local instance, verify diff shows actual differences


xurxodev (changes requested):
thanks @ifoche

This PR fixes a real bug: a missing await on metadataPayloadBuilder.build() in CreatePackageUseCase was storing an unresolved Promise as package contents, makinage empty. The fix itself is correct andminimal. The branch originally also narrowed TransformationRepository.mapPackageTo/mapPackageFrom from generic to a fiMetadataPackage type as an extra safety net a — that part has since been reverted on this branch back to development's fully generic signature, because it broke other legitimate callers once merged with development (see below). What's left to fix ile.

1. Should fix                                                                                                                      - src/domain/packages/usecases/__tests__/Creat doubles don't follow the project'sestablished pattern. The rest of the codebase builds test doubles with ts-mockito (mock/when/instance) plus givenA... helper functions that keep construction out osrc/domain/metadata/builders/__tests__/MetadataPayloadBuilder.spec.tsx, e.g. givenMetadataPayloadBuilderOfCategory). Tnew spec instead hand-rolls every dependency lt inline in beforeEach, and reimplementsStorageClient/DynamicRepositoryFactory collaborators by hand instead of using mock(...)/instance(...). Worth aligning with the existing convention for consistency ng out of sync with how the rest of the suitemocks the same collaborators (DynamicRepositoryFactory, UserRepository, TransformationRepository).

2. Recommendations non blocking                                                                                          
- Worth a short follow-up task to revisit TransformationRepository.mapPackageTo/mapPackageFrom's type safety. The         original commit on this branch narrowed the sackage (Input/Output removed) specifically tomake an unresolved Promise a compile-time error — a reasonable goal given this bug's root cause. After merging development into this branch, that narrowing t legitimately pass non-MetadataPackage shapes: DownloadPayloadUseCase/DownloadPayloadFromSyncRuleUseCase (pass SynchronizationPayload, a wider union) and  TEID2ApiRepository (passes TrackerPostRequest a bug — by teiTransformations-api.spec.ts,which exercises teiTransformations against DHIS2 wire-format data end-to-end). Reverting to the fully generic signature   was the safe choice for this PR's scope, but -leak hole is back open at the type level forany future caller. A follow-up should explore a narrower fix — e.g. a whitelist-style conditional type covering the known valid payload shapes (SynchronizationPayload  than touching this again inside an unrelated PR.

xurxodev:
Hi @ifoche, pushed a commit addressing the "Should fix" point from the review:

- **Tests** (`7edbfe62`): rewrote `CreatePackageUseCase.spec.ts` to use `ts-mockito` (`mock`/`when`/`instance`) with `givenA...` helpers, matching the convention used elsewhere (e.g. `MetadataPayloadBuilder.spec.tsx`), instead of hand-rolled `vi.fn()` doubles and a real `DynamicRepositoryFactory` instance.

  While at it, also simplified the assertions: the original tests verified that internal collaborators were called (`metadataPayloadBuilder.build()`, `transformationRepository.mapPackageTo()`) rather than the use case's actual observable output. Those collaborator-call checks were redundant — if `build()` weren't called, or `mapPackageTo()` were skipped or wrong, the persisted `contents` would already fail to match, since it flows directly into what's saved. Went with 3 tests focused purely on what `execute()` persists (package contents resolved instead of a Promise, provided contents used as-is, module version updated), which is both less coupled to implementation and a closer fit to the bug this PR actually fixes.

All 441 tests pass.

The non-blocking recommendation about `TransformationRepository.mapPackageTo`/`mapPackageFrom` type safety (reverted to fully generic after merging `development`) is left as a follow-up, as originally suggested — it needs a proper design (e.g. a whitelist-style conditional type) to avoid breaking `DownloadPayloadUseCase`/`DownloadPayloadFromSyncRuleUseCase`/`TEID2ApiRepository` again.

xurxodev (approved):

PR #1051, opened 2026-03-04
fix(mapping): correct category option listing in global mapping dialog

ifoche:
## Summary

- Fix duplicate category options in global mapping destination picker — each option now appears exactly once instead of N times (once per parent category)
- Fix mapped selections appearing lost when re-opening the mapping dialog
- Fix automap results being invisible in the selection window

**Root cause**: `MappingDialog` always used `modelFactory("categoryOptions")` which resolved to `CategoryOptionMappedModel` — a model that expands each category option into N composite-ID rows. The fix passes the correct model from `MappingTable` to `MappingDialog` via the config, using `GlobalCategoryOptionModel` (plain IDs, no expansion) for global mapping while preserving composite IDs for nested mapping.

## Related Tasks
- [Category option wrong mapping](https://app.clickup.com/t/869bvjqn0)

## Test plan
- [ ] Run `yarn test` — all 388 tests pass (47 files)
- [ ] Manual: go to Global Mapping, open category options mapping dialog, verify no duplicates
- [ ] Manual: map a category option to a destination, close and re-open dialog, verify selection persists
- [ ] Manual: use automap on category options, open dialog, verify automap result is visible
- [ ] Manual: verify nested category option mapping (within a category combo) still works with composite IDs


xurxodev (approved):
thanks @xurxodev

Tested de fix:

Before
Image

After
Image


This PR fixes global category option mapping, which was broken in three related ways (duplicate destination entries, lost selection on reopen, invisible automap results). The root cause was correctly identified: MappingDialog re-resolved the model via modelFactory(mappingType) instead of reusing the model the table already had

I've seen some recomendation to investigate 

- modelFactory's "first match wins" resolution is fragile beyond this fix, and the fragility isn't fixed — only avoided at this one call site. findClasses("mappingType", ...) uses _.find, so any time more than one class shares the same mappingType the result depends purely on declaration order in mapping.ts. Today there are 3 other such collisions (options, trackedEntityAttributesToDE, aggregatedDataElements); none currently produce a visible bug because either the colliding classes are behaviorally equivalent for the UI, or every call site that matters already passes an explicit model (the same pattern this PR just applied to MappingDialog). But that's incidental, not guaranteed — a future class addition with a modelTransform under one of those mappingTypes could silently reintroduce the same class of bug this PR fixes, with no type error to catch it (MappingDialogConfig.model is optional, so a new caller could simply forget to pass it).
  - A reasonable direction to explore (not implementing here, since it's a larger refactor than this bugfix's scope): either (a) make modelFactory aware of context (global vs. nested) instead of resolving by mappingType alone, or (b) make model required on MappingDialogConfig so every caller is forced to supply it explicitly and modelFactory stops being reachable from this path at all, or (c) assert in modelFactory (e.g. in a test) that no two classes share a mappingType unless they're provably interchangeable for UI purposes.
  - Flagging this as a separate concern for the team/PM to decide whether it's worth a follow-up ticket — it's pre-existing design debt, not something introduced by this PR, and out of scope for this fix.
- No automated test accompanies the fix, consistent with the rest of this component tree (no existing tests for MappingDialog/MappingTable). Worth considering a lightweight regression test if this area gets touched again, but not blocking here given the existing convention.

PR #1055, opened 2026-03-13
[fix]: Compulsory DE operands import error

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869cemv66

### :memo: Implementation
**Problem**: When dataSets are exported, compulsory data element operands may lack a `categoryOptionCombo.id`, causing import failures.
**Solution**: This fix ensures all data element operands in the dataset payload have valid coc id, defaulting to the "default" coc id if not present

### :bookmark_tabs: Others

#869cemv66


src/domain/metadata/builders/MetadataPayloadBuilder.ts
```diff
@@ -379,6 +386,41 @@ export class MetadataPayloadBuilder {
...
+            return dataSets;
+        }
+
+        return dataSets.map(dataSet => {
+            if (!dataSet.compulsoryDataElementOperands) return dataSet;
+
+            const fixedCompulsoryDEOperands = dataSet.compulsoryDataElementOperands.map(operand => {
+                if (operand.categoryOptionCombo?.id) return operand;
+
+                return defaultCategoryOptionCombo
```
gqcorneby: `defaultCategoryOptionCombo` will always be truthy at this point because a guard is in place after declaration. The ternary operation can be simplified into a spread directly

src/domain/metadata/builders/MetadataPayloadBuilder.ts
```diff
@@ -379,6 +386,41 @@ export class MetadataPayloadBuilder {
...
+                    ? {
+                          ...operand,
+                          categoryOptionCombo: { ...operand.categoryOptionCombo, id: defaultCategoryOptionCombo.id },
```
gqcorneby: I don't have full context on the expected shape of categoryOptionCombo, but since we're already guarding against a missing id (`if (operand.categoryOptionCombo?.id)` check above), are there other fields worth preserving via the spread?

src/domain/metadata/builders/MetadataPayloadBuilder.ts
```diff
@@ -379,6 +386,41 @@ export class MetadataPayloadBuilder {
...
+        dataSets: DataSet[]
+    ): Promise<DataSet[]> {
+        const metadataRepository = this.repositoryFactory.metadataRepository(originInstance);
+        const categoryOptionCombos = await metadataRepository.getCategoryOptionCombos({
+            code: { eq: "default" },
+        });
+
+        const defaultCategoryOptionCombo = categoryOptionCombos[0];
+        if (!defaultCategoryOptionCombo) {
+            console.error(
```
gqcorneby, line 388: just confirming: Should this use debug() to stay consistent with the rest of the class's logging pattern, or is console.error intentional here given it's a real error condition?

src/domain/metadata/builders/__tests__/MetadataPayloadBuilder.spec.tsx
```diff
@@ -258,6 +259,7 @@ describe("MetadataPayloadBuilder", () => {
             }
 
             when(mockedMetadataRepository.listAllMetadata(anything())).thenResolve([]);
+            when(mockedMetadataRepository.getCategoryOptionCombos(anything())).thenResolve([]);
```
gqcorneby, line 261: Might be worth adding a test for when getCategoryOptionCombos returns a real combo, verifying that operands with a missing categoryOptionCombo.id get it replaced with the default one.

gqcorneby (commented):
[code-only review] Thanks @deeonwuli! Just some minor comments and clarifications :)

src/domain/metadata/builders/MetadataPayloadBuilder.ts
deeonwuli: `console.error` is intentional here. I'm considering a missing default category option an error here as it could result in the bug this PR is fixing.

src/domain/metadata/builders/MetadataPayloadBuilder.ts
deeonwuli: This fix specifically targets `categoryOptionCombos` that might exist without an id property. The spread preserves any other fields the origin instance included on the object so we're only correcting the id, not reshaping the payload. I have updated the guard to skip operands whose categoryOptionCombo is missing entirely, so the fix now only applies when the object exists but its id is absent.

gqcorneby (approved):
[code-only review] Thanks @deeonwuli!!

xurxodev (changes requested):
Thanks @deeonwuli 

I tried test following the steps in the task using the dataset that indicate the task in our WIDP instance but I can't because it's failing a sync using dataset NTD GNARF(NnhyjiUbcJN)

I did a direct url call and the compulsory data element operands is ok.

Image

This PR asume this bug and patch the payload if categoryOptionCombo.id is empty but what is the reason for this situation in the sync? 

because a direct api url is ok how you can see in the image and this url: https://dev.eyeseetea.com/who-dev-41/api/dataSets/NnhyjiUbcJN?fields=id,name,compulsoryDataElementOperands

In this moment I have the doubt if it's possible fix the problem without patch the payload but in the previous calls. I don't know if you reviewed this possibility, if it's possible the patch is not necessary and the fix should be done before.

Anyway I can't tests because the sync is failing, can you review it?

Image

deeonwuli:
Hi @xurxodev, thanks for the review. You were right that patching the payload was not fixing it at the source.                                           
                                                                                                                                                                                                                                                                                                                                                                                                                                                      
I found that the root cause was `MetadataD2ApiRepository.getMetadata` was calling `/api/metadata` with `defaults=EXCLUDE`. For datasets, this strips the default `categoryOptionCombo.id` from every object in `compulsoryDataElementOperands`.                                                                                                                                                  
  
**Fix**: Instead of correcting operands during payload construction, the fetch itself now preserves those nested default references:

**About the testing blocker (NTD GNARF / 500 error)**: I raised the issue with Arnau in this [ClickUp task](https://app.clickup.com/t/8699f0vdj?comment=90120150220598). This error has not been further explored and has not come up in any of our other projects yet. So it looks unrelated to this PR.

xurxodev (approved):
thanks @deeonwuli 

Now I think we have the correct solution

PR #1059, opened 2026-04-24
fix: replicate rule inclusions

MatiasArriola:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869d1ndau #869d1ndau

### :memo: Implementation

When replicating a rule, the wizard's initialization effect was calling `setDefaultInclusions` on the cloned rule, overwriting the source's sharing/users/orgUnits include-objects-and-references flags with the user's default inclusion config.

- Create any rule with the three inclusion options at `Only references` (sharing, users, organization units)
- Replicate the rule
- Before the fix: inclusion options were not correctly replicated (they were overwritten with defaults)
- After the fix: inclsion options match the source rule


ifoche:
Task linked: [CU-869d1ndau Metadata sync is not replicating all dependencies on rule clone](https://app.clickup.com/t/869d1ndau)

MiquelAdell (approved):

PR #1064, opened 2026-05-14
Basic 2.42 support for EFH

MatiasArriola:
### :pushpin: References

-   **Issue:** https://app.clickup.com/t/869d78d3e #869d78d3e

### :memo: Implementation

-   Upgrade `@eyeseetea/d2-api` from `1.16.0` to `1.21.0` and move the app typings/imports from DHIS `2.36` to `2.40` to support EFH / DHIS 2.42.
-   Refactor D2 model/schema access to use shared helpers (`getApiModel`, `getModelSchema`) so runtime lookups keep working with the newer d2-api typing model.
-   Update current user handling to use top-level `username` and `userRoles` instead of `userCredentials.*`, which is required by newer DHIS versions.
-   Adapt metadata, sync, permissions and package import flows to the new user shape and to the updated d2-api exports.
-   Update tracker/events/TEI repositories to use the new tracker types and keep TEI/event synchronization working with the 2.40 API surface.
-   Add compatibility fixes around metadata handling:
    -   fallback to top-level `userRoles` in metadata transformations
    -   use `code:eq:default` as a temporary workaround when retrieving default metadata objects
    -   keep dataset payload building compatible with default category option combo handling
-   To fetch Event files, use `/tracker/events` instead of `/events` which is deprecated in v36+ and removed in v42
-   Update UI summary/include-exclude screens to resolve model schemas through the new helpers instead of direct `api.models[...]` access.
-   Stabilize tests after the d2-api upgrade by forcing `fetch` through the XHR polyfill so Mirage/Pretender can still intercept requests, and by removing brittle assertions that depended on metadata fetch call order.
-   Add stricter Yarn security settings in `.yarnrc.yml`.

-   N/A

-   Only EFH operations were tested (metadata sync and events sync) - we might need to cover other operations during testing.
-   No update to d2-api v42 yet, I initially upgraded to 1.21 as a transitory step, but basic features seem to work
- ⚠️ This branch drops support for versions < 40. 
  - `userCredentials` handling completely removed (not a fallback)
  - `EventsD2ApiRepository.ts:348-348` now uses the Tracker API - Any sync rule that copies FILE_RESOURCE or IMAGE data values from an older instances might 404 and abort the event sync (I haven't verified minimum version support for the new endpoint).
-  ⚠️ use `code:eq:default` as a temporary workaround when retrieving default metadata objects, to workaround core `identifiable` bug. I haven't verified if `code` is enough for all cases.

### :bookmark_tabs: Others

    No.

    No backend/API repo changes in this PR. This branch only upgrades the app dependency to `@eyeseetea/d2-api@1.21.0`.


Ramon-Jimenez (approved):
Thanks @MatiasArriola , EFH usecase working smoothly on this PR.

PR #1065, opened 2026-05-14
fix: events sync error due to spaces in fields filters

MatiasArriola:
### :pushpin: References

-   **Issue:** Discovered as part of https://app.clickup.com/t/869d78d3e

### :memo: Implementation

- Tested in v42 only, but I assume it can be reproducible in all versions
- The problem is that if any Route url contains unencoded spaces it will fail with a 400 error.

Example: 

GET `http://localhost:8081/dhis2/api/routes/HC1CpN3KuQy/run/api/metadata?fields=id,programType, programTrackedEntityAttributes[trackedEntityAttribute],programStages[id,programStageDataElements[dataElement]]&filter=id:in:[cn4tDPLFfOT]&defaults=EXCLUDE` :

```
{
    "httpStatus": "Bad Request",
    "httpStatusCode": 400,
    "status": "ERROR",
    "message": "Illegal character in query at index 60: http://{ip:port-redacted}/api/metadata?fields=id,programType, programTrackedEntityAttributes[trackedEntityAttribute],programStages[id,programStageDataElements[dataElement]]&filter=id:in:[cn4tDPLFfOT]&defaults=EXCLUDE"
}
```

- After removing the space, it works fine. This was happening in `TEIsPayloadMapperFactory` but looked for other instances of spaces in filters and found only `EventsPayloadMapperFactory`

- Run any events sync

### :bookmark_tabs: Others

No

No

### DIFF
--- a/src/domain/events/mapper/EventsPayloadMapperFactory.ts
+++ b/src/domain/events/mapper/EventsPayloadMapperFactory.ts
@@ -42,7 +42,7 @@ async function getMappedTrackedProgramStages(
     if (destinationProgramStages && destinationProgramStages?.length > 0) {
         const result = await remoteMetadataRepository.getMetadataByIds<ProgramStageRef>(
             destinationProgramStages,
-            "id, program"
+            "id,program"
         );
 
         return result.programStages ?? [];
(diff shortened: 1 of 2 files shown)


xurxodev (approved):
Thanks @MatiasArriola

PR #1067, opened 2026-06-04
[fix]: Keep select-all banner to current type and display cross-type selection notification

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869cz9me0

### :memo: Implementation
Since [#1048](https://github.com/EyeSeeTea/metadata-synchronization/pull/1048) the table receives a flat selection merged across all metadata types. This caused two problems, both fixed here:
- [x] **Select-all banner stuck.** d2-ui compares `selection.length === total`, but the selection carried other-type ids (e.g. 222) while `total` was only the current type's count (e.g. 221), so the "All N selected" banner never resolved. The selection passed to the table is now scoped to the current metadata type's ids.
- [x] **Cross-type selections were invisible.** With the selection scoped down, the user had no feedback that items of other metadata types were still selected. We now show an informational banner line with the count.

https://github.com/user-attachments/assets/f1ae62e6-3736-45d3-8de0-8c919374a741

### Notes to the tester
- [ ] Select all in one metadata type → "All N selected" banner appears and resolves correctly.
- [ ] With items selected in another type, confirm the "N items are selected in other metadata types" line shows.

#869cz9me0


ifoche:
Task linked: [CU-869cz9me0 [metadata-sync] "Select all" notification does not disappear with cross-type selections](https://app.clickup.com/t/869cz9me0)

MiquelAdell (approved):

src/presentation/react/core/components/metadata-table/useSelection.ts
```diff
@@ -0,0 +1,59 @@
...
+    }, [collectionName, selectedIds, compositionRoot, remoteInstance]);
+
+    const idSet = ids.length > 0 ? new Set(ids) : undefined;
+    const selection = (idSet ? selectedIds.filter(id => idSet.has(id)) : selectedIds).map(id => ({
+        id,
+        checked: true,
+        indeterminate: false,
+    }));
+
+    const crossTypeCount = idSet ? selectedIds.filter(id => !idSet.has(id)).length : otherTypeCountFallback;
```
gqcorneby, line 40: `idSet` (line 33) is built from `ids`, but `ids` comes from the `listAll` effect, and `buildListFilters` only forces selected ids into the result when `showOnlySelected` is true. With a search / group / level / lastUpdated filter active, `ids` is restricted to matching rows, so same-type selections that don't match the filter fall outside `idSet` and get counted here as cross-type.

Scenario: in Category, select first 2 items


Search "AMR"


The warning shows "1 items are selected in other metadata types."

gqcorneby (commented):
Thanks @deeonwuli! Just one minor inline comment 👍

**Noting: Pre-existing issue**
While testing the metadata selection, I noticed it's slow, and clicking another item before the previous one finishes processing can drop the earlier selection. The selection-merge logic and row-fetch effect involved here aren't changed by this PR, so it looks like a pre-existing race condition rather than something introduced in this PR.

https://github.com/user-attachments/assets/69e8f223-930e-4318-bea5-94d762f58a1c

@MiquelAdell flagging for visibility. Maybe worth investigating and fixing?

src/presentation/react/core/components/metadata-table/useSelection.ts
MiquelAdell: I have created a specific task for this so I'm closing this PR 
https://app.clickup.com/t/4528615/869dm6r9a

PR #1068, opened 2026-06-04
[fix]: Clear all metadata types when the Clear action is used

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/4528615/869cz9mgn

### :memo: Implementation
"Clear" previously only cleared the current metadata type, selections in other types were silently kept (the cross-type merge from [#1048](https://github.com/EyeSeeTea/metadata-synchronization/pull/1048) re-added them). The fix uses an empty selection as the signal to clear every type: clicking "Clear" is the only thing that empties the selection, because items from other types aren't shown in the current table and so can't be unchecked one by one.

https://github.com/user-attachments/assets/022b1475-7ba0-4c51-a8de-b55063f9861c

### Notes to the tester
- [x] Select items in two metadata types, click **Clear** → all types cleared.
- [x] Select within a single type and clear → behaves as before.

#869cz9mgn


ifoche:
Task linked: [CU-869cz9mgn [metadata-sync] "Clear all" only clears current metadata type, not others](https://app.clickup.com/t/869cz9mgn)

MiquelAdell (approved):

src/presentation/react/core/components/metadata-table/MetadataTable.tsx
```diff
@@ -628,23 +629,33 @@ const MetadataTable: React.FC<MetadataTableProps> = ({
                 parseChildren,
             });
 
-            const mergedSelection = _.uniq([...otherMetadataTypeIds, ...result.included]);
+            const mergedSelection = getMergedSelection({
```
gqcorneby, line 631: this update (similar to line 654) affects the guard below `if (!_.isEqual(stateSelection, mergedSelection))`

`stateSelection` is only updated by `handleTableChange` so clicking clear still does nothing if selection was supplied by the parent. 

step 2 -> select items -> proceed to next step -> go back to step 2 -> select a different category -> clear doesn't work

gqcorneby (changes requested):
Thanks @deeonwuli! Just one inline comment :)

src/presentation/react/core/components/metadata-table/MetadataTable.tsx
deeonwuli: Good catch! Switched the guard to compare selectedIds (the effective selection) instead of stateSelection, so clear now notifies even when the selection comes from the parent.

src/presentation/react/core/components/metadata-table/MetadataTable.tsx
```diff
@@ -605,6 +605,7 @@ const MetadataTable: React.FC<MetadataTableProps> = ({
         const { sorting, pagination, selection } = tableState;
 
         const included = _.reject(selection, { indeterminate: true }).map(({ id }) => id);
+        const isCleared = selection.length === 0;
```
gqcorneby: ⚠ Regression after the `development` merge 
`useSelection` scopes the table's controlled `selection` to the current type, which breaks `isCleared = selection.length === 0` two ways.

Steps to reproduce:
- Select some indicators.
- Switch to the data elements tab (nothing selected there). Wait for it to load.

**1. Clear option no longer visible**
Image

- Go to 2nd page. 
**2. Selection is wiped.**

gqcorneby (changes requested):
Almost there! There's a regression after merging dev

src/presentation/react/core/components/metadata-table/MetadataTable.tsx
```diff
@@ -605,8 +605,8 @@ const MetadataTable: React.FC<MetadataTableProps> = ({
         const { sorting, pagination, selection } = tableState;
 
         const included = _.reject(selection, { indeterminate: true }).map(({ id }) => id);
-
         const [prevMetadataTypeIds, otherMetadataTypeIds] = _.partition(selectedIds, id => ids.includes(id));
+        const isCleared = included.length === 0 && prevMetadataTypeIds.length > 0;
```
gqcorneby: This update resolved the issue I previously reported but this still clears all selection when deselecting to 0 on current category
1. in category type, select an item
2. update type to category combo
3. select one
4. deselect one clears all

Final fix, I think, is to just update to `prevMetadataTypeIds.length > 1`

src/presentation/react/core/components/metadata-table/MetadataTable.tsx
```diff
@@ -662,12 +672,38 @@ const MetadataTable: React.FC<MetadataTableProps> = ({
...
+                                      notifyNewSelection([], []);
+                                  }}
+                                  style={{
+                                      textTransform: "none",
+                                      padding: 0,
+                                      minWidth: "auto",
+                                      verticalAlign: "baseline",
+                                  }}
+                              >
+                                  {i18n.t("Clear")}
```
gqcorneby: Just so the "Clear" and "Clear selection" isn't confusing, maybe we cna udpate this to "Clear all"?

gqcorneby (changes requested):
Thanks @deeonwuli! One last thing and I think we're good :)

gqcorneby (approved):
Thanks @deeonwuli!

PR #1069, opened 2026-06-09
[performance]: Reduce duplicate constant API requests on app load

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869d11cuy

### :memo: Implementation

-    On app load the same `GET /api/constants?...filter=code:eq:MDSYNC_config` request fired multiple times (each returning the same, often empty, result).
-   Root cause: `DynamicRepositoryFactory.getbyInstance` read its created-repository cache with `` `${key}-${tag}` `` but wrote it with `key`, so the cache never hit and `configRepository(instance)` rebuilt a fresh `StorageClientD2Repository` (and `StorageConstantClient`) on every call. The existing per-instance `@cache()` therefore couldn't dedupe across consumers (migrations, settings, store, instance, …).
-   Fixed the factory to read/write the cache with the same key, keyed by `instance.id` — dedupes lookups for the same instance while keeping different target instances isolated (cross-instance sync stays correct).

### :bookmark_tabs: Others

#869d11cuy

### DIFF
--- a/src/data/common/InmemoryCache.ts
+++ b/src/data/common/InmemoryCache.ts
@@ -11,12 +11,22 @@ export class InmemoryCache {
 
     async getOrPromise<T>(cacheKey: string, promise: () => Promise<T>): Promise<T> {
         if (this.cache.has(cacheKey)) {
-            const data = this.cache.get(cacheKey) as T;
-            return Promise.resolve(data);
+            // Cached entry may be an in-flight promise or a resolved value; awaiting handles both.
+            return (await this.cache.get(cacheKey)) as T;
         } else {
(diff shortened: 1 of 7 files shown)


ifoche:
Task linked: [CU-869d11cuy Multiple calls to constants endpoint make rules go slow](https://app.clickup.com/t/869d11cuy)

gqcorneby (approved):
Cool! Thanks @deeonwuli 👍 I did a quick test and requests were reduced ( 5 -> 3 ). Expected since `@cache()` is per-instance and a few `StorageConstantClient` instances still exist on load

PR #1047, opened 2026-02-18
Program indicator sync succeeds but returns no data

anagperal:
### :pushpin: References

-   **Issue:** Closes #869bvm29p [Program indicator sync succeeds but returns no data](https://app.clickup.com/t/869bvm29p)

### :memo: Implementation

-Fix mismatch in download event payload and sync  event payload using `trackedEntities` everywhere

### :video_camera: Screenshots/Screen capture

[Screencast from 2026-02-18 11-58-59.webm](https://github.com/user-attachments/assets/765df01b-8cfc-4082-b8d3-6ba4d58f314a)

### :bookmark_tabs: Others

-   Any change in the [GUI library](https://github.com/EyeSeeTea/d2-ui-components)? If so, what branch/PR?

-   Any change in the [D2 Api](https://github.com/EyeSeeTea/d2-api)? If so, what branch/PR?

PR #1039, opened 2025-11-05
Support for aggregated data exchange

xurxodev:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869avf93v #869avf93v

### :memo: Implementation

- [x] Support ADEX in instances
- [x] Support ADEX in sync rules
- [x] Support dhis2 scheduler for adex

### :fire: Is there anything the reviewer should know to test it?

Limitations:

1- It's not possible to edit an ADEX without assign credentials, for this reason is necessary set credentials to create or update ADEX sync rule
2 - ADEX run failed for datasets, dataElementGroups, etc. The unique models working and enabled are indicators, program indicators and data elements

Scenario to test using play 2.41.6 as origin and play 2.42.3 as destination

1 - In 2.41.6 replicate the user tQSUE8azWFG with all aggregated data exchange authorities

2 - Run analytics 

3 - In 2.42.3 import the next metadata with dataset and data elements with necessary codes 

[adex-play-destination.json](https://github.com/user-attachments/files/23446179/adex-play-destination.json)

4 - Execute Metadata sync against play 2.41.6 using replicated new user credentials

5 - Create a new ADEX instance with url https://play.im.dhis2.org/stable-2-42-3

Screenshot 2025-11-05 at 16 25 29

6 - Create a ADEX events sync rule

Screenshot 2025-11-05 at 16 26 23

Select Sierra Leone org unit

Screenshot 2025-11-05 at 16 26 54

Select doses program indicators: BCG doses, Measles doses, Yellow fever doses

Screenshot 2025-11-05 at 16 27 36

Select this year period

Screenshot 2025-11-05 at 16 27 58

Select monthly aggregation 

Screenshot 2025-11-05 at 16 28 28

Select ADEX instance 

Screenshot 2025-11-05 at 16 29 09

Set credentials

Screenshot 2025-11-05 at 16 29 39

Save

7 - Create a ADEX aggregated sync rule

Screenshot 2025-11-05 at 16 31 39

Select data element: Accute Flaccid Paralysis (Deaths < 5 yrs)

Screenshot 2025-11-05 at 16 33 13

Select indicator: ANC 2 Coverage

Screenshot 2025-11-05 at 16 33 49

Select Sierra Leone org unit
Screenshot 2025-11-05 at 16 34 18

Select this year period

Screenshot 2025-11-05 at 16 34 39

Select monthly aggregation 

Screenshot 2025-11-05 at 16 35 02

Select ADEX instance 

[...]

PR #1070, opened 2026-06-10
[fix]: Prevent dropped selections and fix cross-type count in metadata table

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869dm6r9a

### :memo: Implementation
- [x] **Race condition (dropped selections):** Selection update was gated on an async `getByIds` call, so rapid clicks merged off stale state and overwrote each other. Selection now updates synchronously via `useLatestRef`.
- [x] **Cross-type count overcount:** With an active search/group/level filter, filtered-out same-type selections were miscounted as cross-type. Extended the existing API-based count (already used for org units) to cover this case.

Pre-existing issues flagged by in PR #[1067](https://github.com/EyeSeeTea/metadata-synchronization/pull/1067).

https://github.com/user-attachments/assets/2b35462d-0486-42c2-8d82-19953b72abc4

### :bookmark_tabs: Others

#869dm6r9a

### DIFF
--- a/src/domain/synchronization/usecases/DownloadPayloadFromSyncRuleUseCase.ts
+++ b/src/domain/synchronization/usecases/DownloadPayloadFromSyncRuleUseCase.ts
@@ -59,11 +59,12 @@ export class DownloadPayloadFromSyncRuleUseCase implements UseCase {
 
         const sync: GenericSyncUseCase = this.compositionRoot.sync[rule.type](rule.toBuilder());
 
-        const aggregateDataExchanges = rule.aggregatedDataExchanges?.map(ade => ade.id) || [];
+        // unused
+        // const aggregateDataExchanges = rule.aggregatedDataExchanges?.map(ade => ade.id) || [];
 
-        const payload: SynchronizationPayload = rule.useAggregatedDataExchange
-            ? await this.aggregatedDataExchangeExecutor.getSourceData(aggregateDataExchanges)
-            : await this.buildPayload(rule.type, rule);
+        // const payload: SynchronizationPayload = rule.useAggregatedDataExchange
+        //     ? await this.aggregatedDataExchangeExecutor.getSourceData(aggregateDataExchanges)
+        //     : await this.buildPayload(rule.type, rule);
 
         const date = moment().format("YYYYMMDDHHmm");
 
--- a/src/domain/synchronization/usecases/DownloadPayloadFromSyncRuleUseCase.ts
+++ b/src/domain/synchronization/usecases/DownloadPayloadFromSyncRuleUseCase.ts
@@ -59,13 +59,6 @@ export class DownloadPayloadFromSyncRuleUseCase implements UseCase {
 
         const sync: GenericSyncUseCase = this.compositionRoot.sync[rule.type](rule.toBuilder());
 
-        // unused
-        // const aggregateDataExchanges = rule.aggregatedDataExchanges?.map(ade => ade.id) || [];
-
-        // const payload: SynchronizationPayload = rule.useAggregatedDataExchange
-        //     ? await this.aggregatedDataExchangeExecutor.getSourceData(aggregateDataExchanges)
-        //     : await this.buildPayload(rule.type, rule);
-
         const date = moment().format("YYYYMMDDHHmm");
 
         const mappedData =
--- a/src/presentation/react/core/components/metadata-table/useSelection.ts
+++ b/src/presentation/react/core/components/metadata-table/useSelection.ts
@@ -18,7 +18,7 @@ export function useSelection(
     // When `ids` is not the full id list for the current type (filter active, or collection
     // never populates ids like organisationUnits), set-difference miscounts same-type
     // selections outside the filter as cross-type. Fall back to resolving types via the API.
-    const useApiCrossTypeCount = !idsAreComplete;
+    const shouldFetchCrossTypeData = !idsAreComplete;
     const idCollectionCache = useRef(new Map<string, string | null>());
     useEffect(() => {
         // Invalidate the cache when the resolution context changes (different instance or root).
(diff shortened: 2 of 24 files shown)


ifoche:
Task linked: [CU-869dm6r9a Fix race condition in metadata selection causing dropped selections](https://app.clickup.com/t/869dm6r9a)

src/domain/synchronization/usecases/DownloadPayloadFromSyncRuleUseCase.ts
```diff
@@ -59,11 +59,12 @@ export class DownloadPayloadFromSyncRuleUseCase implements UseCase {
 
         const sync: GenericSyncUseCase = this.compositionRoot.sync[rule.type](rule.toBuilder());
 
-        const aggregateDataExchanges = rule.aggregatedDataExchanges?.map(ade => ade.id) || [];
+        // unused
```
gqcorneby: Looks like Domain and Data layer changes are all just clean up 👍

I just did a quic chec on this but I think we can safely remove it instead of commenting it out.

Initially updated here #1039 then orphaned here #1047
`getSourceData` could also be cleaned up, I think. Can be done in a separate cleanup task

src/presentation/react/core/components/sync-wizard/common/MetadataSelectionStep.tsx
```diff
@@ -146,22 +132,49 @@ export default function MetadataSelectionStep({ syncRule, onChange }: SyncWizard
...
+
+        if (onlyMetadataIds.length === 0) {
             setIdsToIgnore([]);
+            applyTypes(_.uniq(metadataModelsSyncAll));
+            return;
         }
-    }, [compositionRoot.metadata, metadataModelsSyncAll, remoteInstance, syncRule.metadataIds]);
+
+        let cancelled = false;
+        compositionRoot.metadata.getByIds(onlyMetadataIds, remoteInstance, "id").then(metadata => {
```
gqcorneby: Should we add a `.catch` here? This sets the rule's metadata types + aggregation flag, so a silent failure leaves the rule in a stale state. Probably worth surfacing via the existing snackbar like the other effects.

src/presentation/react/core/components/metadata-table/useSelection.ts
```diff
@@ -10,25 +10,35 @@ export function useSelection(
...
             setOtherTypeCountFallback(0);
             return;
         }
 
-        compositionRoot.metadata
-            .getByIds(selectedIds, remoteInstance, "id")
-            .then(pkg => setOtherTypeCountFallback(countPackageOtherType(pkg, collectionName)));
-    }, [collectionName, selectedIds, compositionRoot, remoteInstance]);
+        let cancelled = false;
+        compositionRoot.metadata.getByIds(selectedIds, remoteInstance, "id").then(pkg => {
```
gqcorneby: Whenever a filter is active this re-resolves the whole `selectedIds` set via `getByIds` on every selection change, even though only one id is new per click. I'm not sure how large selections get in normal use here. If it's small it's probably fine, but the cost scales with the total count (server resolution + payload + parsing), not just the request count, so even if ids are < 50 it's doing more work than the one changed id needs.

Is it worth caching type-resolution by id and only resolving newly-added ids? I think there's a similar implementation in `dhis_tally_sheets useDataSets`. Let me know what you think! If it's too much work then maybe it's not wroth the effort.

gqcorneby (commented):
Thanks @deeonwuli! Works much better now without the long delay 👍 

Just a few minor comments

src/presentation/react/core/components/metadata-table/useSelection.ts
deeonwuli: Got it and implemented now, only new IDs are resolved per selection change instead of the full set each time. 👍

src/presentation/react/core/components/metadata-table/useSelection.ts
```diff
@@ -1,44 +1,65 @@
...
-    // not the current type's full id list: never loaded for organisationUnits, and narrowed
-    // to matching rows when a filter (search/group/level/...) is active. In those cases
-    // same-type selections outside the filter would be miscounted as cross-type, so resolve
-    // the selection's types via the API and count items in any collection other than the
-    // current one.
-    const useApiCrossTypeCount = collectionName === "organisationUnits" || filterRestrictsIds;
+    // When `ids` is not the full id list for the current type (filter active, or collection
+    // never populates ids like organisationUnits), set-difference miscounts same-type
+    // selections outside the filter as cross-type. Fall back to resolving types via the API.
+    const useApiCrossTypeCount = !idsAreComplete;
```
gqcorneby: [nit] `use` is kind of a special word that implies a hook? Maybe we can rename it to `resolveCrossTypeViaApi` or something

src/presentation/react/core/components/metadata-table/useSelection.ts
```diff
@@ -1,44 +1,65 @@
...
+
+        if (uncachedIds.length === 0) {
+            setOtherTypeCountFallback(countOtherType());
+            return;
+        }
+
         let cancelled = false;
-        compositionRoot.metadata.getByIds(selectedIds, remoteInstance, "id").then(pkg => {
-            if (!cancelled) setOtherTypeCountFallback(countPackageOtherType(pkg, collectionName));
+        compositionRoot.metadata.getByIds(uncachedIds, remoteInstance, "id").then(pkg => {
```
gqcorneby: [nit] pkg is a bit cryptic. Maybe `metadataByType`?

gqcorneby (approved):
Thanks @deeonwuli! Just some nit-picks but all good!

gqcorneby (approved):
All good!

adrianq:
@deeonwuli Can you please merge with the development branch, please?

PR #1247, opened 2026-06-18
[fix]: Always build scheduler artifact in `yarn build`

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869dqwwjt

### :memo: Implementation

-   Recent releases were missing the scheduler artifact (`metadata-synchronization-server.zip`) because `yarn build` only produced the front-end variant zips and never invoked `yarn build-scheduler`. It had to be generated manually from the release commit.
-   Wire `build-scheduler` into `build()` in `scripts/run.ts`, running it once after the variant loop, so `yarn build` always emits the scheduler artifact and releases include it.
-   Placed the call inside `build()` (rather than appending `&& yarn build-scheduler` to the `build` npm script) so it runs exactly once regardless of the selected variant and doesn't break positional-arg passing.
-   Updated the README to note that `yarn build` now also produces the scheduler zip.

-   Run `yarn build` (or `yarn build core-app`) and confirm `metadata-synchronization-server.zip` is produced alongside the variant zips.
-   Verify the bundle is present: `unzip -l metadata-synchronization-server.zip | grep index.js`, and that the temp `metadata-synchronization-server/` dir is cleaned up afterwards.
-   Note: under Yarn 4, the existing `build-scheduler` script uses a bare `$npm_package_name`, which Yarn's portable shell mis-parses (`Unbound variable "npm_package_name-server"`). If the build fails on that, the script needs `${npm_package_name}` (braces) instead.

#869dqwwjt

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,7 +177,10 @@ To build a given front-end:
 $ yarn build [all|core-app|data-metadata-app|module-package-app|modules-list|package-exporter|msf-aggregate-data-app|sp-emergency-responses|wmr]
 ```
 
-To build the scheduler:
+`yarn build` already builds the scheduler and emits the `metadata-synchronization-server.zip` file
+alongside the front-end zips, so it is always included in releases.
+
+To build the scheduler on its own:
 
 ```
 $ yarn build-scheduler
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
         "build:variant": "vite build && cp -r i18n icon.png build",
         "preview": "vite preview",
         "prebuild": "yarn localize && yarn test",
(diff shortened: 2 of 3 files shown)


xurxodev (approved):
Thanks @deeonwuli 

Tested with yarn build under Yarn 4 (via Corepack/nvm) — all 8 variants built and zipped correctly, yarn build-scheduler ran automatically at the end without manual intervention, no Unbound variable error, metadata-synchronization-server.zip was produced with index.js inside, and the temp metadata-synchronization-server/ dir was cleaned up afterward. Works as described.

One non-blocking note @adrianq: since build-scheduler now runs on every yarn build, it also runs for single-variant builds (e.g. yarn build core-app), adding ~17s even for local/dev builds that previously didn't need it. Worth keeping in mind if that overhead matters for the dev loop, but not a reason to block this fix.

PR #1248, opened 2026-06-22
[fix]: Scheduler failing on DHIS2 2.43 (userCredentials removed from /me)

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/4528615/869dqwtvv

### :memo: Implementation
**Root cause:** DHIS2 version 2.43 removed the `userCredentials` wrapper from `/api/me`; `username`/`userRoles` are now top-level. Code reading `currentUser.userCredentials.userRoles`/`.username` threw `Cannot read properties of undefined (reading 'userRoles')`. In the scheduler this surfaced only as `[ERROR] scheduler - {}` because the catch blocks serialized the error with `JSON.stringify(error)`, which returns `{}` for `Error` objects.

- [x] Added a `getErrorMessage` helper (`src/utils/error.ts`) and used it in the scheduler catch blocks (`cli.ts`, `SchedulerCLI.ts`) so errors log with their real message/stack instead of `{}`. Also applied it to previously swallowed dataStore read errors (`StorageDataStoreClient`).
- [x] Updated all current-user (`/me`) reads to support both pre-2.43 and 2.43+ shapes via a fallback (`userCredentials?.x ?? x`): `permissions.ts` (`getUserInfo`, `getUserRoles`), `UserD2ApiRepository.getCurrent`, `GenericSyncUseCase`, and the package list/import components (routed through the user use case).

-   Run the scheduler against a DHIS2 2.43 instance that has a scheduler-enabled sync rule. Before this change it fails right after "Loading synchronization rules from remote server"; after, rules load and schedule normally.

#869dqwtvv

### DIFF
--- a/src/data/storage/StorageDataStoreClient.ts
+++ b/src/data/storage/StorageDataStoreClient.ts
@@ -8,6 +8,7 @@ import { getD2APiFromInstance } from "../../utils/d2-utils";
 import { DataStorageType } from "../../domain/storage-client-config/entities/StorageConfig";
 import { Future, FutureData } from "../../domain/common/entities/Future";
 import { apiToFuture } from "../common/utils/api-futures";
+import { getErrorMessage } from "../../utils/error";
 
 export const dataStoreNamespace = "metadata-synchronization";
 
@@ -41,7 +42,7 @@ export class StorageDataStoreClient extends StorageClient {
             const value = await this.dataStore.get<T>(key).getData();
             return value;
         } catch (error: any) {
-            console.error(error);
+            console.error(`Error reading dataStore key "${key}": ${getErrorMessage(error)}`);
             return undefined;
         }
     }
@@ -50,7 +51,7 @@ export class StorageDataStoreClient extends StorageClient {
         return Future.fromPromise(this.dataStore.get<T>(key).getData())
             .map(value => value)
             .mapError(error => {
-                console.error(error);
+                console.error(`Error reading dataStore key "${key}": ${getErrorMessage(error)}`);
                 return error;
             });
     }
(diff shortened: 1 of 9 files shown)


MiquelAdell:
@idelcano could you please provide a functional review of that one?
thanks!

gqcorneby:
Hi @deeonwuli! Heads up that this branch is behind development and conflicts with the d2-api upgrade that merged last week (#1064), which touched the same userCredentials handling.

Related PR: https://github.com/EyeSeeTea/metadata-synchronization/pull/1064

Could you resolve conflicts and give it a quick check on 2.43 before I finalize? Worth a look at #1064 first, there's real overlap, so the scope here likely narrows to the logging improvement. 🙏

xurxodev (commented):
@deeonwuli @MiquelAdell @gqcorneby 

I think this PR is not necessary and it can be closed. The user credentials is a 2.42 problem not 2.43 and it was solved by @MatiasArriola in this PR https://github.com/EyeSeeTea/metadata-synchronization/pull/1064

MiquelAdell:
thank you all for your comments.
@MatiasArriola can you confirm that this PR #1248 is no longer needed after #1064 ?

MatiasArriola:
> thank you all for your comments. @MatiasArriola can you confirm that this PR #1248 is no longer needed after #1064 ?

`userCredentials` usage was already removed in the referenced PR. But this PR also improves error handling / reporting in the scheduler. Maybe we can still benefit from that part (I guess it surfaced only because of userCredentials-related errors, but will help in the future for any other class of error). 

The diff is outdated since this branch needs to incorporate changes from development and resolve the conflicts (which are in fact the previous userCredentials implementation). So if we want to keep part of this PR, it would stop being "Scheduler failing on DHIS2 2.43 (userCredentials removed from /me)" and become something like "Improve error reporting in Scheduler"

PR #1252, opened 2026-07-07
Support internal (same-instance) Aggregate Data Exchange

anagperal:
### :pushpin: References

-   **Issue:** Closes #869dzkuyv [Support internal (same-instance) Aggregate Data Exchange](https://app.clickup.com/t/4528615/869dzkuyv)
-  https://github.com/EyeSeeTea/metadata-synchronization/pull/1039

### :memo: Implementation

Adds the ability to run an Aggregated Data Exchange against the **current instance itself** (source = destination), alongside the existing external exchanges.

#### Feature
- When creating an *aggregated data exchange* instance, a new **Exchange target type** selector lets you choose **External** (existing behavior) or **Internal**.
- For an **Internal** exchange the endpoint URL is hidden and automatically set to the current (local) instance URL — no manual URL entry.
- Only **one** Internal exchange instance can exist.
- Internal exchanges require **no credentials**.
- **Exchange target type** and **instance type** are **immutable after creation**: enforced on save and reflected in the UI (the selectors are disabled when editing).
- Internal exchanges use the **UID** identifier scheme; external exchanges keep **CODE** (same code across instances, different UID).
- When building a synchronization rule of type *Aggregated data exchange* (Aggregated Data Sync or Event Sync):
  - Selecting the **internal** instance needs no credentials and shows no credentials step.
  - The credentials dialog and the **Set Credentials** button only appear when at least one **external** target is selected.
  - In a mixed selection, the internal target is listed as “no credentials required” next to the external ones.

#### Bugs fixed
- The exchange target type was not persisted when an instance was first created (it only appeared after editing it once); it is now saved from creation.
- The credentials dialog wrongly appeared for an **internal-only** selection — a failure that only reproduced when the app ran **installed in the instance** (not on localhost), because an internal exchange shares the local URL. Internal targets are now seeded reliably regardless of environment or sync type.
- The event-only “generate new UIDs” behavior was being triggered incorrectly for internal ADEX (which shares the local URL); it is now scoped to non-ADEX event syncs.
- Creating a rule could fail with a DHIS2 error when the exchange source request name exceeded 50 characters; the request name is now truncated while the exchange name stays full.
- The Data Exchange app showed an “unsupported identifier scheme” warning for internal exchanges; internal now uses UID consistently, clearing it.
- External exchanges using an API token were being wrongly flagged as missing credentials; the completeness check now honors the auth type.
- **Download JSON** failed for Aggregated Data Exchange rules — both internal and external, and from both the wizard Summary step and the rules list — with a generic download error. The preview is now built directly from the rule’s selection (metadata, org units, periods) and no longer depends on the exchange already existing in DHIS2 or on a fixed identifier scheme.
- Fixed and extracted translatable strings (i18n) for the new UI.

**Create internal ADEX**

[test create internal.webm](https://github.com/user-attachments/assets/db21bb1f-8e60-45ab-b5da-9575cdd896b9)

**Events ADEX**

https://github.com/user-attachments/assets/6e20670a-6855-495b-a26a-ada461d3a5b0

test program indicator

test PI to DE done4
test PI to DE done3
test PI to DE done2
test PI to DE done1

**Aggregated ADEX**

[test aggregated- DE to DE.webm](https://github.com/user-attachments/assets/bfc37d16-459a-40f5-8ae6-da702572c6bb)

DE souce:

test agg DE5
test agg DE4
test agg DE3
test agg DE2
test agg DE1

Result:

test agg DE done5
test agg DE done4
test agg DE done3
test agg DE done2
test agg DE done1

- This is the dataset with the target data element from the program indicator:

[dataset-with-DE-from-PI-metadata.json](https://github.com/user-attachments/files/29743445/dataset-with-DE-from-PI-metadata.json)

- It is necesary to set the aggregateExportDataElement field  (the uuid of the target DE) in the program indicator metadata (via API or metadata import, as it may not be available in the UI yet only in Metadata Management App in version 42).

image

### :bookmark_tabs: Others

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2026-06-16T15:11:31.847Z\n"
-"PO-Revision-Date: 2026-06-16T15:11:31.854Z\n"
+"POT-Creation-Date: 2026-07-07T09:14:53.473Z\n"
+"PO-Revision-Date: 2026-07-07T09:14:53.473Z\n"
 
 msgid ""
 "THIS NEW RELEASE INCLUDES SHARING SETTINGS PER INSTANCES. FOR THIS VERSION "
@@ -37,6 +37,19 @@ msgstr "Program not found"
 
 msgid "An instance with this name already exists"
 msgstr "An instance with this name already exists"
+
+msgid "The instance type cannot be changed after the instance is created"
+msgstr "The instance type cannot be changed after the instance is created"
+
+msgid "The exchange target type cannot be changed after the instance is created"
+msgstr "The exchange target type cannot be changed after the instance is created"
+
+msgid ""
+"An internal Aggregated Data Exchange instance already exists: "
+"{{instanceName}}"
+msgstr ""
+"An internal Aggregated Data Exchange instance already exists: "
+"{{instanceName}}"
 
 msgid ""
 "An Aggregated Data Exchange instance with this URL already exists: "
@@ -182,6 +195,12 @@ msgstr "Ok"
 
 msgid "Cancel"
 msgstr "Cancel"
+
+msgid "Instance: {{name}}"
+msgstr "Instance: {{name}}"
+
+msgid "Internal exchange — no credentials required."
+msgstr "Internal exchange — no credentials required."
 
 msgid "Authentication Scheme (*)"
 msgstr "Authentication Scheme (*)"
@@ -1958,6 +1977,12 @@ msgstr "API Token"
 msgid "Aggregated Data Exchange"
 msgstr "Aggregated Data Exchange"
 
+msgid "External"
+msgstr "External"
+
+msgid "Internal"
+msgstr "Internal"
+
 msgid "Please fix the issues before testing the connection"
 msgstr "Please fix the issues before testing the connection"
 
@@ -1966,6 +1991,9 @@ msgstr "Connected successfully to instance"
 
 msgid "Server name (*)"
 msgstr "Server name (*)"
+
+msgid "Exchange target type (*)"
+msgstr "Exchange target type (*)"
 
 msgid "URL endpoint (*)"
 msgstr "URL endpoint (*)"
--- a/i18n/es.po
+++ b/i18n/es.po
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: i18next-conv\n"
-"POT-Creation-Date: 2026-06-16T15:11:31.847Z\n"
+"POT-Creation-Date: 2026-07-06T15:09:17.686Z\n"
 "PO-Revision-Date: 2020-07-10T06:53:30.625Z\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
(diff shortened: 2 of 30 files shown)


xurxodev (changes requested):
thanks @anagperal 

Functionally works ok but I've detected some issues in the code to improve

1. Should fix

- RuleAggregatedDataExchange.isMissingCredentials doesn't check username (src/domain/rules/value-object/RuleAggregatedDataExchange.ts:35-38). For http-basic, it only checks !this.target.password. If password is set but username is empty/missing, this getter returns false (not missing), so SynchronizationRule.validate() (SynchronizationRule.ts:958) and the wizard's hasInstancesWithoutCredentials check (Steps.ts:63) both consider the rule complete — but DHIS2 will reject basic auth with an empty username at execution time. Suggest: !this.target.username || !this.target.password for the http-basic branch. No test currently covers "password set, username missing" — the existing suite jumps straight from "both set" to "password absent" (RuleAggregatedDataExchange.spec.ts:252-259).

2. Recommendations non blocking
 - buildAggregatedDataExchangePayload silently builhen instance is undefined(src/data/rules/utils/buildAggregatedDataExchangePayload.ts:19,25,42-43). If the target instance was deleted after the rule wacreated, this produces type: "EXTERNAL", api: { urudly. Not introduced by this PR, but worth afollow-up.                                                                                                                    
 - A shared useLocalInstance() hook would remove dulerplate. GeneralInfoForm.tsx:60-66 adds a 4thnear-identical copy of compositionRoot.instances.getById("LOCAL").then(...match({success, error})) — the same pattern already exists in InstanceSelectionStep.tsx:64-73, CreateP3, and NewPackageDialog.tsx:79-88. Two of the fourcopies now live in the same instance-creation/sync-wizard feature area touched by this PR, so it's a reasonable moment to extract a shared hook instead of adding a fourth copy.

anagperal:
> thanks @anagperal
> 
> Functionally works ok but I've detected some issues in the code to improve
> 
> 1. Should fix
> 
> * RuleAggregatedDataExchange.isMissingCredentials doesn't check username (src/domain/rules/value-object/RuleAggregatedDataExchange.ts:35-38). For http-basic, it only checks !this.target.password. If password is set but username is empty/missing, this getter returns false (not missing), so SynchronizationRule.validate() (SynchronizationRule.ts:958) and the wizard's hasInstancesWithoutCredentials check (Steps.ts:63) both consider the rule complete — but DHIS2 will reject basic auth with an empty username at execution time. Suggest: !this.target.username || !this.target.password for the http-basic branch. No test currently covers "password set, username missing" — the existing suite jumps straight from "both set" to "password absent" (RuleAggregatedDataExchange.spec.ts:252-259).
> 
> 2. Recommendations non blocking
> 
> * buildAggregatedDataExchangePayload silently builhen instance is undefined(src/data/rules/utils/buildAggregatedDataExchangePayload.ts:19,25,42-43). If the target instance was deleted after the rule wacreated, this produces type: "EXTERNAL", api: { urudly. Not introduced by this PR, but worth afollow-up.
> * A shared useLocalInstance() hook would remove dulerplate. GeneralInfoForm.tsx:60-66 adds a 4thnear-identical copy of compositionRoot.instances.getById("LOCAL").then(...match({success, error})) — the same pattern already exists in InstanceSelectionStep.tsx:64-73, CreateP3, and NewPackageDialog.tsx:79-88. Two of the fourcopies now live in the same instance-creation/sync-wizard feature area touched by this PR, so it's a reasonable moment to extract a shared hook instead of adding a fourth copy.

@xurxodev thanks!! all these changes are implemented now.

xurxodev (approved):
Thanks @anagperal

bhavananarayanan (approved):

PR #1264, opened 2026-08-05
[Security] resolve dependency vulnerabilities

anagperal:
### :pushpin: References

- **Issue:** Closes #869e9vzg7 [analyze dependency issues](https://app.clickup.com/t/4528615/869e9vzg7)
- **Findings on this branch:** [open code-scanning alerts](https://github.com/EyeSeeTea/metadata-synchronization/security/code-scanning?query=is%3Aopen+branch%3Afix%2Fdependency-vulnerabilities+) — filter by tool `OWASP Dependency-Track (yarn4)` and `is:open`.
- **Documentation:** findings without a reachable fix are recorded in `RESOLUTIONS.md`, each with the reason and the condition that would let it be closed.
- **Reference implementation:** [dhis2-app-skeleton#120](https://github.com/EyeSeeTea/dhis2-app-skeleton/pull/120) — open and not approved. Used as a reference and re-verified against this tree, not copied.
- Process notes are in the [internal wiki](https://github.com/EyeSeeTea/internal-know-how/wiki/Resolving-Dependency-Vulnerabilities) (draft, for team discussion).

### :memo: Implementation

A dedicated remediation pass over the open Dependency-Track findings on this branch. Every finding
with a published fix that is reachable without a toolchain migration is resolved here; the ones that
are not are documented in `RESOLUTIONS.md` with the reason and the condition that would let them be
closed.

#### Fixed without adding any pin

**Re-resolution inside existing ranges — no manifest change.** The largest part of this PR is a
lockfile refresh. For most findings the parent's declared range already admitted the patched release
and the lockfile had simply gone stale, so `yarn up -R` reached the fix with nothing added to
`package.json`:

| Package                                                | Now resolves to        |
| ------------------------------------------------------ | ---------------------- |
| `vitest` (and `@vitest/ui`, which requests it exactly) | 3.2.7                  |
| `brace-expansion`                                      | 1.1.18 / 2.1.4 / 5.0.9 |
| `fast-uri`                                             | 3.1.5                  |
| `form-data`                                            | 2.5.6 / 3.0.5 / 4.0.6  |
| `ip-address`                                           | 10.3.1                 |
| `js-yaml`                                              | 3.15.0 / 4.3.0         |
| `@babel/plugin-transform-modules-systemjs`             | 7.29.7                 |
| `postcss`                                              | 8.5.24                 |
| `tar`                                                  | 7.5.22                 |
| `vite` (the vitest chain)                              | 7.3.6                  |
| `ws`                                                   | 7.5.13 / 8.21.1        |

Note that the existing `tar`, `request/form-data` and `vitest/vite` entries needed no edit — they
were already written as ranges and were doing their job; only the lockfile was behind.

⚠️ The `js-yaml` row closed the advisory it was aimed at, but a **new** one (`GHSA-5p4m-2wfm-xmqj`,
2026-08-06) lands on the very versions it resolved to. It is tracked under _Remaining findings_ and
closes in a follow-up commit — the same shape of fix, blocked for now by the same install-age policy.

#### Decayed pins bumped to `^` ranges

**Two exact-version constraints converted to ranges.** Both had decayed into the problem they were
added to prevent:

- `axios` was held at exactly `1.16.0`. A later advisory affects `>= 1.15.2, < 1.18.0`, and because
  the constraint was an exact version no patch could ever be selected. Now `^1.18.0`.
- `qs` was held at exactly `6.14.2`. A later advisory affects `>= 6.11.1, <= 6.15.1`, so the
  constraint was holding the tree _at_ the vulnerable version rather than above it. Now `^6.15.3`.

⚠️ Neither could simply be deleted: `@eyeseetea/d2-api` requests both at older exact versions, so
removing the constraint resolves them _downwards_. They change shape, not existence.

#### Scoped pins added

**Six scoped resolutions**, each for a path where the parent's range cannot reach a patched release
and the parent itself has no release that fixes it:

| Resolution                            | Why the range cannot reach the fix                                        |
| ------------------------------------- | ------------------------------------------------------------------------- |
| `react-linkify/linkify-it: ^5.0.2`    | `react-linkify@1.0.0-alpha` requests `^2.0.3`; the fix is on the 5.x line |
| `styled-components/postcss: ^8.5.18`  | `styled-components@6.1.8` requests `postcss` at an exact version          |
| `external-editor/tmp: ^0.2.6`         | `external-editor@3.1.0` requests `^0.0.33`                                |
| `isomorphic-fetch/node-fetch: ^2.6.7` | `isomorphic-fetch@2.2.1` requests `^1.0.1`; the fix is on 2.x             |
| `@dhis2/cli-app-scripts/vite: ^6.4.3` | the bundled vite is on 5.x, and there is no fix on that line              |
| `i18next-conv/node-gettext: ^3.0.1`   | `i18next-conv@9.2.1` requests `^2.0.0`; the fix is on 3.x                 |

> The `node-gettext` one is worth calling out: its advisory records **no patched version at all**, so a
> tool reading only that field reports it as a dead end. Its affected range is `<= 3.0.0` and 3.0.1 is
> published and outside it. Comparing the affected range against the published version list, rather
> than trusting the patched-version field, is what found it.

#### Removed

**One constraint removed.** `minimatch@npm:10.2.4/brace-expansion` still matched a descriptor, but
made no difference to the resolved version: `minimatch@10.2.4` requests `^5.0.2` and `10.2.5`
requests `^5.0.5`, and both already admit every published patch on that line. Removing it and
re-installing left the resolved version unchanged. It was also the highest-decay-risk shape in the
file — a versioned-parent entry that stops matching silently when the parent patch-bumps.

#### Medium-severity pass

Handled in a second pass, and mostly the same story: re-resolution inside the declared ranges cleared
`ajv` to 6.15.0, `bn.js` to 4.12.5 and 5.2.5, and `yaml` to 1.10.3, with no manifest change.

`@babel/core` was pinned as a **direct devDependency** at the exact `7.15.8`, which held all three of
its lines below the patch. Reopened to `^7.29.6`; the tree now collapses onto a single 7.29.7.

Three more scoped resolutions, all inside the build-only `@dhis2/cli-app-scripts` chain:

| Resolution                                   | Why the range cannot reach the fix                                       |
| -------------------------------------------- | ------------------------------------------------------------------------ |
| `http-proxy-agent/@tootallnate/once: ^2.0.1` | `http-proxy-agent@4.0.1` requests `1`; the fix for that line is on 2.0.1 |
| `request/tough-cookie: ^4.1.3`               | `request@2.88.2` requests `~2.5.0`; the fix is on the 4.x line           |
| `package-json/got: ^11.8.5`                  | `package-json@6.5.0` requests `^9.6.0`; the fix is on 11.8.5             |

All three cross a major into packages that are deprecated or written against the older API, so each
was checked by loading its consumer rather than by `yarn install` alone — see _Notes for the reviewer_.

> **A fourth was attempted and had no effect, so it is not here.** `vite@npm:^4.0.0/esbuild: ^0.25.0`
> was tried against `esbuild@0.18.20`. The build passed, but the lockfile came back byte-identical and
> `vite@4.5.14` still received 0.18.20 — even though the sibling entry `vite@npm:^4.0.0/rollup` binds
> correctly. Had the green build been taken as the signal, this would have shipped as a fix that fixes
> nothing. It was removed rather than left in as decoration, and the attempt is recorded in
> `RESOLUTIONS.md` so nobody repeats it.

#### Corrections to `RESOLUTIONS.md`

Two recorded statements were no longer true and would have misled the next person:

- The `minimatch` entries gave **eslint 9** as their drop-when condition. That is wrong.
  `eslint@9.39.5` still declares `minimatch ^3.1.5` and reaches the same line again through
  `@eslint/eslintrc@^3.3.6`, which declares `^3.1.5` too. `eslint@10.8.0` is the first release that
  drops `@eslint/eslintrc` and moves to `minimatch ^10.2.5`. Anyone acting on the old condition would
  have removed ten load-bearing entries and reintroduced the finding.
- The `brace-expansion` entry claimed the 1.x and 2.x lines were already patched natively. They were
  not.

The file also gains a floor-versus-fixture convention, a section for findings with no fix available,
and a decay check for constraints whose removal changes nothing.

#### Remaining findings

Every finding still reported on this branch, at any severity, and why each one is still there. Check
the [current alerts](../../security/code-scanning?query=is%3Aopen+branch%3Afix%2Fdependency-vulnerabilities)
rather than this list: it is accurate on the day it was written and the scanner is the authority.

This table covers the Dependency-Track findings, which is what this PR is scoped to. The alert list
also carries two CodeQL `js/redos` findings against the application's own source
(`src/domain/packages/validator/validations/ALL-MQ-18/validate_ALL-MQ-18.ts`). Those are not
dependency issues, are not addressed here, and need their own change.

| Advisory                                                              | Component                                    | Severity | Why it is still here                                                                                                                              |
| --------------------------------------------------------------------- | -------------------------------------------- | -------- | ------------------------------------------------------------------------------------------------------------------------------------------------- |
| `GHSA-5p4m-2wfm-xmqj`                                                 | `js-yaml@3.15.0`, `@4.3.0`                   | high     | Published 2026-08-06, after this branch's analysis. Fix exists and both ranges admit it — blocked only by the install-age policy, see below       |
| `GHSA-gv7w-rqvm-qjhr`                                                 | `esbuild@0.18.20`, `@0.25.11`, `@0.25.12`    | high     | **Advisory withdrawn 2026-06-17.** Not a defect; to be dismissed rather than remediated                                                           |
| `GHSA-w5hq-g745-h8pq`                                                 | `uuid@3.4.0`                                 | high     | No published release fixes it — see below                                                                                                         |
| `GHSA-fx2h-pf6j-xcff` · `GHSA-c27g-q93r-2cwf`                         | `vite@4.5.14`                                | high     | No patched release on the 4.x line — see below                                                                                                    |
| `GHSA-p5wg-g6qr-c7cg`                                                 | `eslint@8.57.1`                              | medium   | **Advisory withdrawn 2026-02-03.** Same treatment as the `esbuild` one                                                                            |
| `GHSA-p8p7-x288-28g6`                                                 | `request@2.88.2`                             | medium   | The advisory records no patched version, and 2.88.2 is the last release ever published — see below                                                |
| `GHSA-848j-6mx2-7j84`                                                 | `elliptic@6.6.1`                             | medium   | The advisory covers every published version; 6.6.1 is the latest                                                                                  |
| `GHSA-67mh-4wv8-2f99`                                                 | `esbuild@0.18.20`                            | medium   | Patched at 0.25.0, which vite 4's `^0.18.10` cannot reach — same migration as the `vite` findings                                                 |
| `GHSA-v6wh-96g9-6wx3` · `GHSA-4w7w-66w2-5vf9` · `GHSA-93m4-6634-74q7` | `vite@4.5.14`                                | medium   | Same 4.x line, same migration                                                                                                                     |
| `GHSA-g4jq-h2w9-997c` · `GHSA-jqfw-vq24-v9c3`                         | `vite@4.5.14`                                | low      | Same                                                                                                                                              |

**The `js-yaml` finding is the only one here with a reachable fix.** `GHSA-5p4m-2wfm-xmqj` was
published on 2026-08-06, after the rest of this work was done, and reports quadratic CPU consumption
in `!!omap` resolution — the 5.x fix for `CVE-2026-59870` was never backported to the 3.x and 4.x
lines. Unlike everything else in this table it is not a dead end: 3.15.1 and 4.3.1 were published on
07-31, and both declared ranges (`^3.13.1` / `^3.14.1` and `^4.1.0`) already admit them, so it is pure
re-resolution with no manifest change. It is held back only by the install-age policy described under
_Notes for the reviewer_, and lands in a follow-up commit on this branch. Every consumer is toolchain —
`eslint`, `@eslint/eslintrc`, `@istanbuljs/load-nyc-config` and `depcheck` — parsing the project's own
config files at build time, so the practical exposure to an untrusted-YAML denial of service is low.

**The `vite` findings** are against the application's own `vite@4.5.14`. There is no patched release
anywhere on the 4.x line — the two highs are patched at 6.4.3 and 5.4.9 — so neither re-resolution nor
a scoped resolution can reach a fix. The only remediation is moving the application to vite 6 or
later, which changes the build configuration and needs its own testing. Deliberately left for the
separate vite migration rather than turned into a toolchain migration here. These are dev-server
surfaces and none reaches the production bundle. The `esbuild@0.18.20` finding rides on the same
migration: it is vite 4's own esbuild, and a scoped resolution against it was tried and had no effect.

**The `uuid` and `request` findings** share a chain and a single exit. Both are reached through
`@dhis2/cli-app-scripts` → `@dhis2/cli-helpers-engine` → `request@2.88.2`. For `uuid`, no published
release satisfies both the advisory and the `uuid/v4` subpath that `request` imports, since that
subpath was removed in v7 — forcing the patched version fails at load with
`ERR_PACKAGE_PATH_NOT_EXPORTED`. For `request` itself, the advisory records no patched version and the
published version list confirms it: 2.88.2 is the last release, and the package has been deprecated
since 2020. Both are build-only, and both close the day `@dhis2/cli-helpers-engine` drops `request`.

**Two of the remaining findings are withdrawn advisories** — `esbuild` and `eslint`. Neither describes
a real defect, and the upgrades they appear to call for correct nothing. Worth noting that the
`eslint` one is scored medium, which is why it survived a first pass filtered to critical and high:
withdrawn advisories can sit below the gate's threshold and go unexamined for longer than the ones
above it.

#### Documentation

Everything above except `js-yaml` is recorded in `RESOLUTIONS.md` under _Known findings with no fix
available_, each with its chain and the condition that would let it be closed. `js-yaml` is not,
because it does not belong there: it has a reachable fix and, being pure re-resolution, will leave no
trace in the manifest once it lands.

### :fire: Notes for the reviewer

Run against the consumer of every package whose version changed, not just `yarn install`:

| Check                      | Result                                                                                                                                                                                                                                       |
| -------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `yarn install --immutable` | clean, lockfile reproducible                                                                                                                                                                                                                 |
| `tsc --noEmit`             | clean                                                                                                                                                                                                                                        |
| `yarn lint`                | clean                                                                                                                                                                                                                                        |
| `yarn test`                | 465 tests across 57 files, all passing                                                                                                                                                                                                       |
| `yarn localize`            | extracts the same 840 strings, no change to `i18n/en.pot` or the `.po` files beyond their generation timestamps                                                                                                                              |
| `yarn build`               | the full build, not just `build:variant`: every variant, its manifest and zip, plus `build-scheduler`, which is a separate `ncc` bundle of the scheduler CLI and the only place several of the changed runtime packages get bundled for Node |

Seven changes cross a major or move a package that a parent pins exactly, so each was checked against
the component that uses it rather than assumed:

| Change                                                 | Check                                                                                                                                                                                                                                   |
| ------------------------------------------------------ | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **`linkify-it` 2 → 5**                                 | still exports a callable CJS function with the same `.tlds()`, `.match()` and `.test()` surface, and `<Linkify>` renders the expected `<a href>` for both URLs and `mailto:` addresses. `<Linkify>` is used by the store-creation pages |
| **`postcss` under `styled-components`**                | renders through `ServerStyleSheet` emitting the declared properties and the nested `:hover` rule                                                                                                                                        |
| **`node-fetch` 1 → 2**                                 | `isomorphic-fetch` loads and still installs a global `fetch`                                                                                                                                                                            |
| **`node-gettext` 2 → 3** and **the pinned `vite`**     | exercised through `yarn localize`, which is the command that actually uses them                                                                                                                                                         |
| **`tough-cookie` 2 → 4 under `request`**               | `request.jar()` still creates a jar, `request.cookie()` parses, and `getCookieString` round-trips the value. `request` is deprecated and written against the tough-cookie 2 API, so this is the one worth checking rather than assuming |
| **`got` 9 → 11 under `package-json`**                  | `package-json` and `update-notifier` both load                                                                                                                                                                                          |
| **`@tootallnate/once` 1 → 2 under `http-proxy-agent`** | `HttpProxyAgent` instantiates                                                                                                                                                                                                           |

Two more things to know while reviewing:

- `.yarnrc.yml` sets `npmMinimalAgeGate: 7d`, so releases published within the last week cannot be
  selected. `brace-expansion` 1.1.18 / 2.1.4 / 5.0.9 and `fast-uri` 3.1.5 were held by it and landed
  as soon as it opened; `js-yaml` 3.15.1 / 4.3.1 is still inside the window and lands the same way.
  The gate is a supply-chain control and was not weakened to work around any of them.
- `cross-spawn@5.1.0` is reported by a local `yarn npm audit` but not by the Dependency-Track
  analysis. It is left alone deliberately, with the reasoning recorded under _Future improvements_.

No user-facing change — dependencies and configuration only.


xurxodev (approved):
thanks @anagperal

